PR #4, opened 2015-05-22
Only attempt to build X11 on Linux

larsbergstrom:
x11 panics the build if it is built on anything other than linux-targeting-linux, due to pkg_config issues with xlib.

[Review on Reviewable](https://reviewable.io/reviews/ecoal95/rust-offscreen-rendering-context/4)


larsbergstrom:
Not needed with @Daggerbot's changes to x11 version 1.1.1.

PR #7, opened 2015-06-03
Use platform dependencies and remove Cargo.lock.

metajack:
This updates the Cargo.toml to use platform dependency support which is now working in Cargo. This also removes the Cargo.lock file which is not supposed to be included in libraries.


emilio:
Nice! The Cargo.lock was there since when servo used a "outdated" rustc version. Thanks for the PR, hopefully it'll make servo build ultra-fast :)

PR #10, opened 2015-06-09
Porting to raspberry pi2.

Adenilson:
Please consider merging this as it is required to compile servo in Raspberry pi.


emilio:
Nice, thanks! :)

PR #17, opened 2015-06-15
Use gleam and cgl from crates.io

SimonSapin:


emilio:
@SimonSapin maybe using wildcards could be better, to make less painful updates for servo?

SimonSapin:
In Cargo, `gleam = "0.1"` is equivalent to `gleam = "^0.1"` which is equivalent to `gleam = ">=0.1 <0.2"`. So `cargo update` will automatically pick up 0.1.1, 0.1.2 etc when we publish them. If we follow SemVer, 0.2 means that there was a breaking change and we might not want to update automatically. I think this is what we want.

https://users.rust-lang.org/t/should-be-used-for-cargo-dependencies/313 has some more discussion.

emilio:
Didn't know that, thanks! :)

PR #23, opened 2015-07-14
Allow serialization of WebGL context attributes via `serde`.

pcwalton:
This will allow us to perform the multiprocess split.

r? @glennw


glennw:
r+

PR #24, opened 2015-07-17
Native surfaces now store their size

mrobinson:
We no longer have to pass size as an argument to some methods.


mrobinson:
@glennw @pcwalton r?

glennw:
r+

PR #33, opened 2015-10-15
Allow failure on glutin platform and minor nits

emilio:
We should probably change `Err(&str)` to another type of error...


emilio:
May you take a quick look at this and tell me if it compiles fine on windows @vvuk? :)

(It definitely should, but I have no way to easily test it actually, and our CI is broken because of aster)

emilio:
Seems like CI is broken again due to aster.

I could confirm the build and tests passing on linux and mac, and tested the code used in windows and unimplemented  platforms.

I also changed the feature test, that was working previously just because I was lucky and the vector I was creating contained the memory from old data (I don't even know how I wanted back in the day to get the texture pixels with `TexImage2D` instead of `GetTexImage`).

Android-specific code shouldn't be affected here, so I'm going to merge this.

jdm:
Good find on the feature test problem :)

PR #36, opened 2015-11-01
platform/glx: Check for a possible `XOpenDisplay` error

emilio:
Could that be the cause of https://github.com/servo/servo/issues/7931?

This check should be here before anyway. I find it really extrange that
this fails, but it seems to happen on the buildbots...


emilio:
Skipping CI because of aster again (sigh).

PR #43, opened 2015-11-30
Fix use of glXChooseFBConfig.

glennw:
Based on https://www.opengl.org/sdk/docs/man2/xhtml/glXChooseFBConfig.xml
the last parameter of the attributes array must be None (which == 0).

Confusingly, this is different from GLX_NONE, which is non-zero.

This fixes webgl context creation on some drivers, in particular nVidia proprietary drivers but probably others too.


emilio:
Wow, good catch.

I found this `GLX_NONE`/`None` behaviour completely surprising, and this explains Servo's buildbots crashing... Probably Mesa drivers do ignore invalid attributes while nvidia's reject them...

Thanks! :)

PR #46, opened 2015-12-09
Merge webrender branch

emilio:
May you review it @jdm and/or @glennw?

It's the branch used in webrender with fixes for egl, restored support for OSX, and cleaned up a bit:
- I removed the texture_surface feature that has been unused since always
- And also the windows glutin port since it's not capable of managing current contexts, etc. I'll try to adapt it in the future, but for now no WebGL in windows is fine I guess.
- And proper CI again! \o/

It also supports using an EGL backend from linux (though only via renderbuffer for now, since some gl errors appear with texture formats).


jdm:
I'll need to defer to @glennw or do a lot of studying.

glennw:
@ecoal95 I haven't looked at this yet - but does it contain any changes that rely on the servo/webrender branch? Or are they changes that will work with servo/master right now?

emilio:
This will work with Servo master with minor changes (just passing the
wanted `NativeGLContext` backend).

Nothing that relies on `servo/webrender`, `webrender`, nor
`webrender_traits`.

emilio:
Glenn approved this changes on IRC a few days ago, so I'm merging them and will make the servo PR ASAP.

PR #47, opened 2015-12-22
platform/egl: Fix EGL texture implementation

emilio:
The EGL test failures where provoked by two things:
- GLES 2.0 deprecates `gl::TEXTURE_2D` as a valid parameter to
  `gl::enable`, which was provoking an unexpected INVALID_ENUM error.
- GLES 2.0 doesn't have `glGetTexImage`, so to test that writing to a
  texture actually works we have to bind it to another FBO. We do it on
  the sharing test.

Also, added a no-op feature to easily run egl tests on linux. Too bad we can't test it on Travis :(

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,4 +22,7 @@ os:
 
 script:
   - cargo build --verbose
+  # We cannot test egl nor glx in Travis' linux
+  # - cargo test --verbose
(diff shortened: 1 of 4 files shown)


emilio:
Merging it directly since it only affects with the removal of one LoC of non-tests.

PR #48, opened 2016-01-09
New GLLimits struct to access GL "limit values".

simartin:
Implement suggestion in code review for my fix for servo issue #9097 at https://reviewable.io/reviews/servo/servo/9102#-K6p6P67OPgZjXKhbiD8

### DIFF
--- a/src/gl_context.rs
+++ b/src/gl_context.rs
@@ -6,6 +6,7 @@ use NativeGLContextMethods;
 use GLContextAttributes;
 use GLContextCapabilities;
 use GLFormats;
+use GLLimits;
 use DrawBuffer;
 use ColorAttachmentType;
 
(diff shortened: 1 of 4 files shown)


src/gl_context.rs
```diff
@@ -61,6 +65,7 @@ impl<Native> GLContext<Native>
         let mut context = try!(Self::create(shared_with));
 
         context.formats = GLFormats::detect(&attributes);
+        context.limits = GLLimits::detect();
```
emilio: There should be no need to re-detect the limits here.

The formats are needed because the new attributes don't have to be the old ones, but the limits should always be the same.

src/gl_limits.rs
```diff
@@ -0,0 +1,15 @@
+use gleam::gl::types::GLint;
+use gleam::gl;
+
+#[derive(Deserialize, Serialize)]
```
emilio: Probably want to derive `Clone` too here, to send it around without calling `detect()` again.

emilio:
Left a few comments, but this looks good to me :)

The CI failures are due to: https://github.com/rust-lang/rust/issues/30713

simartin:
Thanks, PR updated accordingly

emilio:
Great! LGTM

PR #52, opened 2016-04-06
Update gl_generator to 0.5

jdm:


emilio:
@jdm Travis complained, so I updated your patch with gl_generator's API changes.

If travis passes (it should), I'll merge it and publish it right away.

emilio:
Don't merge this, it panics when calling a glx functions due to they being not loaded, I'm checking if this is expected in https://github.com/bjz/gl-rs/issues/403.

emilio:
This is ready to merge, I used `GlobalGenerator` instead of `StaticGenerator`, and everything keeps working just fine now :)

PR #53, opened 2016-04-26
armv7 support

fabricedesre:


Cargo.toml
```diff
@@ -40,6 +40,10 @@ features = ["xlib"]
 version = "2.3.0"
 features = ["xlib"]
 
+[target.armv7-unknown-linux-gnueabihf.dependencies.x11]
```
jdm: It's legal to do

```
[target.cfg(target_os = "linux").dependencies]
x11 = { version = "2.3.0", features = ["xlib"] }
```

and that would allow us to remove all of these duplicate entries, I'm pretty sure.

fabricedesre: Cool, i'll try that. And maybe I'll update all the other repos where I added the armv7 section...

emilio:
Looks good, also needs a version bump :P

I'd prefer to use the `cfg()` syntax, but if it's blocking or something I can merge it right away and do the changes myself.

SimonSapin:
`lib.rs` uses `cfg(target_os="linux")` and `cfg(target_os="macos")`. ("unix" applies to both.)

emilio:
Y es, xlib should not be used on ~~Marcos~~ Mac OS.

fabricedesre:
Updated, sorry.

emilio:
LGTM, will merge and publish after the travis run :)

Actually don't be sorry, thanks for doing this :)

PR #55, opened 2016-05-12
Add max texture size members to GLLimits

dlrobertson:
As discussed on IRC. Add `max_tex_size` and `max_cube_map_size` to GLLimits. I'll then switch on `target` to select the appropriate max.

Let me know if this is not what you had in mind.

---

This change is [Reviewable](https://reviewable.io/reviews/emilio/rust-offscreen-rendering-context/55)

### DIFF
--- a/src/gl_limits.rs
+++ b/src/gl_limits.rs
@@ -4,12 +4,16 @@ use gleam::gl;
 #[cfg_attr(feature="serde_serialization", derive(Serialize, Deserialize))]
 pub struct GLLimits {
     pub max_vertex_attribs: u32,
+    pub max_tex_size: u32,
+    pub max_cube_map_tex_size: u32
 }
 
 impl GLLimits {
     pub fn detect() -> GLLimits {
         GLLimits {
             max_vertex_attribs: gl::get_integer_v(gl::MAX_VERTEX_ATTRIBS) as u32,
+            max_tex_size: gl::get_integer_v(gl::MAX_TEXTURE_SIZE) as u32,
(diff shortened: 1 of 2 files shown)


src/gl_limits.rs
```diff
@@ -4,12 +4,16 @@ use gleam::gl;
 #[cfg_attr(feature="serde_serialization", derive(Serialize, Deserialize))]
 pub struct GLLimits {
     pub max_vertex_attribs: u32,
+    pub max_tex_size: i32,
+    pub max_cube_map_size: i32
```
emilio: Could you call this `max_cube_map_tex_size`? I think that way is clearer

emilio:
Yes, this looks great, thanks! :-)

Could you please bump the Cargo.toml version and address the other comment?

dlrobertson:
Good point. No problem!

src/gl_limits.rs
```diff
@@ -4,12 +4,16 @@ use gleam::gl;
 #[cfg_attr(feature="serde_serialization", derive(Serialize, Deserialize))]
 pub struct GLLimits {
     pub max_vertex_attribs: u32,
+    pub max_tex_size: i32,
```
emilio: Arrgh, missed this previously, sorry! This should be unsigned (texture sizes are guaranteed to be positive).

dlrobertson: Ah, true. I originally had this as unsigned, but changed it because `level`, `width`, and `height` are signed, but at the time of comparison they are guaranteed to be positive as well. I'll make the update.

emilio:
Thanks! :)

PR #56, opened 2016-06-11
Remove dependency on serde_macros

nox:


emilio:
Cherrypicked with a version bump in master, but what's the reasoning for the change?

nox:
I'm trying to make webrender not rely on serde_macros to rely on less unstable code.

emilio:
Sounds good :)

PR #57, opened 2016-06-14
Make the serde support behave better

nox:
The previous implementation could produce illegal JSON when used
with serde_json.


kvark:
@nox what were the illegal JSONs out of curiosity?

PR #60, opened 2016-08-12
Update serde to 0.8

nox:


nox:
https://travis-ci.org/emilio/rust-offscreen-rendering-context/builds/151776928

PR #62, opened 2016-08-27
Clamp the drawbuffer size to a minnimum, add more tests, and cleanups.

emilio:
The tests in the first round were unsuccessful attempts to reproduce https://github.com/servo/servo/issues/12320.

The second commit fixes that.

r? @jdm


jdm:
Makes sense!

PR #63, opened 2016-09-19
Unset the texture/rb bindings after we use them.

anholt:
Fixes webgl conformance's default-texture.html.


emilio:
LGTM, thanks!

emilio:
Published right now.

PR #64, opened 2016-09-19
Implement WGL

MortimerGoro:
WGL implementation, now the servo webgl triangle demo runs on windows ;)

### DIFF
--- a/src/platform/with_wgl/native_gl_context.rs
+++ b/src/platform/with_wgl/native_gl_context.rs
@@ -1,6 +1,6 @@
 use platform::NativeGLContextMethods;
 use std::ffi::CString;
-use std::os::raw::{c_void};
+use std::os::raw::c_void;
 use std::ptr;
 use std::sync::{Once, ONCE_INIT};
 
@@ -19,11 +19,10 @@ lazy_static! {
     static ref GL_LIB: Option<HMODULEWrapper>  = {
         let p = unsafe{kernel32::LoadLibraryA(b"opengl32.dll\0".as_ptr() as *const _)};
         if p.is_null() {
-            debug!("opengl32.dll not found!");
+            error!("WGL: opengl32.dll not found!");
             None
-        }
-        else {
-            debug!("opengl32.dll LOADED!");
+        } else {
+            debug!("WGL: opengl32.dll loaded!");
             Some(HMODULEWrapper(p))
         }
     };
@@ -32,16 +31,14 @@ lazy_static! {
 static LOAD_GL: Once = ONCE_INIT;
 pub fn load_gl() {
     LOAD_GL.call_once(|| {
-        gl::load_with(|s| {
-            NativeGLContext::get_proc_address(s) as *const _
-        });
+        gl::load_with(|s| NativeGLContext::get_proc_address(s) as *const _);
     });
 }
 
 pub struct NativeGLContext {
-    pub render_ctx: winapi::HGLRC,
-    pub device_ctx:winapi::HDC, //
-    pub weak: bool,
+    render_ctx: winapi::HGLRC,
+    device_ctx: winapi::HDC,
+    weak: bool,
 }
 
 impl Drop for NativeGLContext {
@@ -57,30 +54,42 @@ impl Drop for NativeGLContext {
     }
 }
 
-pub struct NativeGLContextHandle(pub winapi::HGLRC, pub winapi::HDC);
+pub struct NativeGLContextHandle(winapi::HGLRC, winapi::HDC);
 unsafe impl Send for NativeGLContextHandle {}
 
 impl NativeGLContextMethods for NativeGLContext {
     type Handle = NativeGLContextHandle;
 
     fn get_proc_address(addr: &str) -> *const () {
-        let addr = CString::new(addr.as_bytes()).unwrap().as_ptr();
+        let addr = CString::new(addr.as_bytes()).unwrap();
+        let addr = addr.as_ptr();
         unsafe {
             let p = wgl::GetProcAddress(addr) as *const _;
-            if !p.is_null() { return p; }
+            if !p.is_null() {
+                return p;
+            }
             match *GL_LIB {
-                Some(ref lib) =>  kernel32::GetProcAddress(lib.0, addr) as *const _,
-                None => ptr::null_mut()
+                Some(ref lib) => kernel32::GetProcAddress(lib.0, addr) as *const _,
+                None => ptr::null_mut(),
             }
         }
 
     }
 
     fn create_shared(with: Option<&Self::Handle>) -> Result<NativeGLContext, &'static str> {
-        match unsafe{super::utils::create_offscreen(with, &WGLAttributes::default())}{
-            Ok(ctx) => {
-                //wglGetProcAddress only works in the presence of a valid GL context
-                //OpenGL functions must be loaded after the first context is created
+        let render_ctx = match with {
+            Some(ref handle) => handle.0,
+            None => ptr::null_mut(),
+        };
+        match unsafe { super::utils::create_offscreen(render_ctx, &WGLAttributes::default()) } {
+            Ok(ref res) => {
+                // wglGetProcAddress only works in the presence of a valid GL context
+                // OpenGL functions must be loaded after the first context is created
+                let ctx = NativeGLContext {
+                    render_ctx: res.0,
+                    device_ctx: res.1,
+                    weak: false,
+                };
                 ctx.make_current().unwrap();
                 load_gl();
                 Ok(ctx)
@@ -101,40 +110,40 @@ impl NativeGLContextMethods for NativeGLContext {
             Some(NativeGLContext {
                 render_ctx: handle.0,
                 device_ctx: handle.1,
-                weak: true
+                weak: true,
             })
-        }
-        else {
+        } else {
             None
         }
     }
 
     fn current_handle() -> Option<Self::Handle> {
-        let handle = unsafe{ wgl::GetCurrentContext()};
+        let handle = unsafe { wgl::GetCurrentContext() };
         if !handle.is_null() {
-            let hdc = unsafe { wgl::GetCurrentDC()};
+            let hdc = unsafe { wgl::GetCurrentDC() };
             Some(NativeGLContextHandle(handle as winapi::HGLRC, hdc as winapi::HDC))
-        }
-        else {
+        } else {
             None
         }
     }
 
     fn make_current(&self) -> Result<(), &'static str> {
-         if unsafe {wgl::MakeCurrent(self.device_ctx as * const _, self.render_ctx as *const _) != 0 } {
-            Ok(())
-        } else {
-            Err("wgl::makeCurrent failed")
+        unsafe {
+            if wgl::MakeCurrent(self.device_ctx as *const _, self.render_ctx as *const _) != 0 {
+                Ok(())
+            } else {
+                Err("wgl::makeCurrent failed")
+            }
         }
     }
 
     fn unbind(&self) -> Result<(), &'static str> {
-        if self.is_current() {
-            unsafe {wgl::MakeCurrent(ptr::null_mut(),ptr::null_mut() );}
-            Ok(())
-        }
-        else {
-            Err("gwl::MakeCurrent (on unbind)")
+        unsafe {
+            if self.is_current() && wgl::MakeCurrent(ptr::null_mut(), ptr::null_mut()) == 0 {
+                Err("gwl::MakeCurrent (on unbind)")
+            } else {
+                Ok(())
+            }
         }
     }
 
--- a/src/platform/with_wgl/native_gl_context.rs
+++ b/src/platform/with_wgl/native_gl_context.rs
@@ -118,12 +118,12 @@ impl NativeGLContextMethods for NativeGLContext {
     }
 
     fn current_handle() -> Option<Self::Handle> {
-        let handle = unsafe { wgl::GetCurrentContext() };
-        if !handle.is_null() {
-            let hdc = unsafe { wgl::GetCurrentDC() };
-            Some(NativeGLContextHandle(handle as winapi::HGLRC, hdc as winapi::HDC))
-        } else {
+        let ctx = unsafe { wgl::GetCurrentContext() };
+        let hdc = unsafe { wgl::GetCurrentDC() };
+        if ctx.is_null() || hdc.is_null() {
             None
+        } else {
+            Some(NativeGLContextHandle(ctx as winapi::HGLRC, hdc as winapi::HDC))
         }
     }
 
--- a/src/platform/with_wgl/native_gl_context.rs
+++ b/src/platform/with_wgl/native_gl_context.rs
@@ -2,22 +2,21 @@ use platform::NativeGLContextMethods;
 use std::ffi::CString;
 use std::os::raw::c_void;
 use std::ptr;
-use std::sync::{Once, ONCE_INIT};
 
 use winapi;
 use user32;
 use kernel32;
 use super::wgl;
 use super::wgl_attributes::*;
-use gleam::gl;
+use super::utils;
 
-/// Wrapper to satisfy `Sync`.
+// Wrappers to satisfy `Sync`.
 struct HMODULEWrapper(winapi::HMODULE);
 unsafe impl Sync for HMODULEWrapper {}
 
 lazy_static! {
     static ref GL_LIB: Option<HMODULEWrapper>  = {
-        let p = unsafe{kernel32::LoadLibraryA(b"opengl32.dll\0".as_ptr() as *const _)};
+        let p = unsafe { kernel32::LoadLibraryA(b"opengl32.dll\0".as_ptr() as *const _) };
         if p.is_null() {
             error!("WGL: opengl32.dll not found!");
             None
@@ -26,13 +25,23 @@ lazy_static! {
             Some(HMODULEWrapper(p))
         }
     };
-}
 
-static LOAD_GL: Once = ONCE_INIT;
-pub fn load_gl() {
-    LOAD_GL.call_once(|| {
-        gl::load_with(|s| NativeGLContext::get_proc_address(s) as *const _);
-    });
+    static ref PROC_ADDR_CTX: Option<NativeGLContext> = {
+        match unsafe { utils::create_offscreen(ptr::null_mut(), &WGLAttributes::default()) } {
+            Ok(ref res) => {
+                let ctx = NativeGLContext {
+                    render_ctx: res.0,
+                    device_ctx: res.1,
+                    weak: false,
+                };
+                Some(ctx)
+            }
+            Err(s) => {
+                error!("Error creating GetProcAddress helper context: {}", s);
+                None
+            }
+        }
+    };
 }
 
 pub struct NativeGLContext {
@@ -54,8 +63,12 @@ impl Drop for NativeGLContext {
     }
 }
 
+unsafe impl Send for NativeGLContext {}
+unsafe impl Sync for NativeGLContext {}
+
 pub struct NativeGLContextHandle(winapi::HGLRC, winapi::HDC);
 unsafe impl Send for NativeGLContextHandle {}
+unsafe impl Sync for NativeGLContextHandle {}
 
 impl NativeGLContextMethods for NativeGLContext {
     type Handle = NativeGLContextHandle;
@@ -64,10 +77,27 @@ impl NativeGLContextMethods for NativeGLContext {
         let addr = CString::new(addr.as_bytes()).unwrap();
         let addr = addr.as_ptr();
         unsafe {
+
+            if wgl::GetCurrentContext().is_null() {
+                // wglGetProcAddress only works in the presence of a valid GL context
+                // We use a dummy ctx when the caller calls this function without a valid GL context
+                if let Some(ref ctx) = *PROC_ADDR_CTX {
+                    if ctx.make_current().is_err() {
+                        return ptr::null_mut();
+                    }
+                } else {
+                    return ptr::null_mut();
+                }
+            }
+
             let p = wgl::GetProcAddress(addr) as *const _;
             if !p.is_null() {
                 return p;
             }
+            // wglGetProcAddress​ doesn't return function pointers for some legacy functions,
+            // (the ones coming from OpenGL 1.1)
+            // These functions are exported by the opengl32.dll itself,
+            // so we have to fallback to kernel32 getProcAddress if wglGetProcAddress​ return null
             match *GL_LIB {
                 Some(ref lib) => kernel32::GetProcAddress(lib.0, addr) as *const _,
                 None => ptr::null_mut(),
@@ -81,17 +111,15 @@ impl NativeGLContextMethods for NativeGLContext {
             Some(ref handle) => handle.0,
             None => ptr::null_mut(),
         };
-        match unsafe { super::utils::create_offscreen(render_ctx, &WGLAttributes::default()) } {
+        match unsafe { utils::create_offscreen(render_ctx, &WGLAttributes::default()) } {
             Ok(ref res) => {
-                // wglGetProcAddress only works in the presence of a valid GL context
-                // OpenGL functions must be loaded after the first context is created
+
                 let ctx = NativeGLContext {
                     render_ctx: res.0,
                     device_ctx: res.1,
                     weak: false,
                 };
                 ctx.make_current().unwrap();
-                load_gl();
                 Ok(ctx)
             }
             Err(s) => {
--- a/src/platform/with_wgl/native_gl_context.rs
+++ b/src/platform/with_wgl/native_gl_context.rs
@@ -1,0 +1,144 @@
+use platform::NativeGLContextMethods;
+use std::ffi::CString;
+use std::os::raw::{c_void};
+use std::ptr;
+use std::sync::{Once, ONCE_INIT};
+
+use winapi;
+use user32;
+use kernel32;
+use super::wgl;
+use super::wgl_attributes::*;
+use gleam::gl;
+
+/// Wrapper to satisfy `Sync`.
+struct HMODULEWrapper(winapi::HMODULE);
+unsafe impl Sync for HMODULEWrapper {}
+
+lazy_static! {
+    static ref GL_LIB: Option<HMODULEWrapper>  = {
+        let p = unsafe{kernel32::LoadLibraryA(b"opengl32.dll\0".as_ptr() as *const _)};
+        if p.is_null() {
+            debug!("opengl32.dll not found!");
+            None
+        }
+        else {
+            debug!("opengl32.dll LOADED!");
+            Some(HMODULEWrapper(p))
+        }
+    };
+}
+
+static LOAD_GL: Once = ONCE_INIT;
+pub fn load_gl() {
+    LOAD_GL.call_once(|| {
+        gl::load_with(|s| {
+            NativeGLContext::get_proc_address(s) as *const _
+        });
+    });
+}
+
+pub struct NativeGLContext {
+    pub render_ctx: winapi::HGLRC,
+    pub device_ctx:winapi::HDC, //
+    pub weak: bool,
+}
+
+impl Drop for NativeGLContext {
+    fn drop(&mut self) {
+        unsafe {
+            if !self.weak {
+                wgl::DeleteContext(self.render_ctx as *const _);
+                let window = user32::WindowFromDC(self.device_ctx);
+                user32::ReleaseDC(window, self.device_ctx);
+                user32::DestroyWindow(window);
+            }
+        }
+    }
+}
+
+pub struct NativeGLContextHandle(pub winapi::HGLRC, pub winapi::HDC);
+unsafe impl Send for NativeGLContextHandle {}
+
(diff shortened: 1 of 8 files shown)


src/platform/with_wgl/native_gl_context.rs
```diff
@@ -0,0 +1,144 @@
...
+
+/// Wrapper to satisfy `Sync`.
+struct HMODULEWrapper(winapi::HMODULE);
+unsafe impl Sync for HMODULEWrapper {}
+
+lazy_static! {
+    static ref GL_LIB: Option<HMODULEWrapper>  = {
+        let p = unsafe{kernel32::LoadLibraryA(b"opengl32.dll\0".as_ptr() as *const _)};
+        if p.is_null() {
+            debug!("opengl32.dll not found!");
```
emilio: nit: `error!` here will make it appear on the console, and seems more likely what we want.

src/platform/with_wgl/native_gl_context.rs
```diff
@@ -0,0 +1,144 @@
...
+use super::wgl_attributes::*;
+use gleam::gl;
+
+/// Wrapper to satisfy `Sync`.
+struct HMODULEWrapper(winapi::HMODULE);
+unsafe impl Sync for HMODULEWrapper {}
+
+lazy_static! {
+    static ref GL_LIB: Option<HMODULEWrapper>  = {
+        let p = unsafe{kernel32::LoadLibraryA(b"opengl32.dll\0".as_ptr() as *const _)};
```
emilio: nit: `unsafe {`, and `as *const _) };`.

src/platform/with_wgl/native_gl_context.rs
```diff
@@ -0,0 +1,144 @@
...
+    LOAD_GL.call_once(|| {
+        gl::load_with(|s| {
+            NativeGLContext::get_proc_address(s) as *const _
+        });
+    });
+}
+
+pub struct NativeGLContext {
+    pub render_ctx: winapi::HGLRC,
+    pub device_ctx:winapi::HDC, //
```
emilio: nit: space after the colon. Also, either do comment, or remove the two slashes.

Please make the members private, since the intention is making `NativeGLContext` cross-platform. Feel free to add accessor methods if needed.

src/platform/with_wgl/native_gl_context.rs
```diff
@@ -0,0 +1,144 @@
...
+                wgl::DeleteContext(self.render_ctx as *const _);
+                let window = user32::WindowFromDC(self.device_ctx);
+                user32::ReleaseDC(window, self.device_ctx);
+                user32::DestroyWindow(window);
+            }
+        }
+    }
+}
+
+pub struct NativeGLContextHandle(pub winapi::HGLRC, pub winapi::HDC);
```
emilio: Same here regarding `pub`.

src/platform/with_wgl/native_gl_context.rs
```diff
@@ -0,0 +1,144 @@
...
+impl NativeGLContextMethods for NativeGLContext {
+    type Handle = NativeGLContextHandle;
+
+    fn get_proc_address(addr: &str) -> *const () {
+        let addr = CString::new(addr.as_bytes()).unwrap().as_ptr();
+        unsafe {
+            let p = wgl::GetProcAddress(addr) as *const _;
+            if !p.is_null() { return p; }
+            match *GL_LIB {
+                Some(ref lib) =>  kernel32::GetProcAddress(lib.0, addr) as *const _,
```
emilio: hmm... I'm worried we can end up mixing symbols from taken from `GL_LIB` and from wgl. May it be a problem?

If it is, please only call `wgl::GetProcAddress` when it's `None`, if it's not, please write a comment saying why :)

src/platform/with_wgl/native_gl_context.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+pub struct NativeGLContextHandle(pub winapi::HGLRC, pub winapi::HDC);
+unsafe impl Send for NativeGLContextHandle {}
+
+impl NativeGLContextMethods for NativeGLContext {
+    type Handle = NativeGLContextHandle;
+
+    fn get_proc_address(addr: &str) -> *const () {
+        let addr = CString::new(addr.as_bytes()).unwrap().as_ptr();
```
emilio: Please keep one variable for `CString::new(addr.as_bytes()).unwrap()`, and other for `<whatever>.as_ptr()`. 

`CString` may allocate a buffer, and depending on when the compiler executes the destructor of `CString` you could be using it after free. I'm not sure if it can happen in practice, but better safe than sorry.

src/platform/with_wgl/native_gl_context.rs
```diff
@@ -0,0 +1,144 @@
...
+        }
+        else {
+            debug!("opengl32.dll LOADED!");
+            Some(HMODULEWrapper(p))
+        }
+    };
+}
+
+static LOAD_GL: Once = ONCE_INIT;
+pub fn load_gl() {
```
emilio: Hm... I think we shouldn't be calling this under the hood. it should be the callers responsibility. Also, shouldn't be public.

src/platform/with_wgl/native_gl_context.rs
```diff
@@ -0,0 +1,144 @@
...
+                None => ptr::null_mut()
+            }
+        }
+
+    }
+
+    fn create_shared(with: Option<&Self::Handle>) -> Result<NativeGLContext, &'static str> {
+        match unsafe{super::utils::create_offscreen(with, &WGLAttributes::default())}{
+            Ok(ctx) => {
+                //wglGetProcAddress only works in the presence of a valid GL context
```
emilio: nit: space after the comment slashes.

src/platform/with_wgl/native_gl_context.rs
```diff
@@ -0,0 +1,144 @@
...
+            match *GL_LIB {
+                Some(ref lib) =>  kernel32::GetProcAddress(lib.0, addr) as *const _,
+                None => ptr::null_mut()
+            }
+        }
+
+    }
+
+    fn create_shared(with: Option<&Self::Handle>) -> Result<NativeGLContext, &'static str> {
+        match unsafe{super::utils::create_offscreen(with, &WGLAttributes::default())}{
```
emilio: nit: spaces between braces, just as before.

src/platform/with_wgl/native_gl_context.rs
```diff
@@ -0,0 +1,144 @@
...
+        let addr = CString::new(addr.as_bytes()).unwrap().as_ptr();
+        unsafe {
+            let p = wgl::GetProcAddress(addr) as *const _;
+            if !p.is_null() { return p; }
+            match *GL_LIB {
+                Some(ref lib) =>  kernel32::GetProcAddress(lib.0, addr) as *const _,
+                None => ptr::null_mut()
+            }
+        }
+
```
emilio: nit: extra newline

src/platform/with_wgl/native_gl_context.rs
```diff
@@ -0,0 +1,144 @@
...
+                weak: true
+            })
+        }
+        else {
+            None
+        }
+    }
+
+    fn current_handle() -> Option<Self::Handle> {
+        let handle = unsafe{ wgl::GetCurrentContext()};
```
emilio: nit: spacing in braces, here and in the rest of the file.

src/platform/with_wgl/native_gl_context.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    fn current() -> Option<Self> {
+        if let Some(handle) = Self::current_handle() {
+            Some(NativeGLContext {
+                render_ctx: handle.0,
+                device_ctx: handle.1,
+                weak: true
+            })
+        }
```
emilio: nit: rust style is `} else {`.

src/platform/with_wgl/native_gl_context.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    fn unbind(&self) -> Result<(), &'static str> {
+        if self.is_current() {
+            unsafe {wgl::MakeCurrent(ptr::null_mut(),ptr::null_mut() );}
+            Ok(())
+        }
+        else {
+            Err("gwl::MakeCurrent (on unbind)")
```
emilio: We definitely shouldn't return an error if the current context was already unbound, the error condition should be `wgl::MakeCurrent` failing.

emilio (changes requested):
Super exciting to see this!

I need to run and I still haven't made it to the end. A lot of formatting nits and a few questions, but I still need to reach to the core of this PR, I'll try to do it over the rest of this week.

Thanks!

emilio:
(Sorry, misclick)

src/platform/with_wgl/native_gl_context.rs
MortimerGoro: I thought about that too but we might need an API change to avoid loading the symbols under the hood in WGL. The callers can't load the symbols because wglGetProcAddress only works in the presence of a valid GL context. On the other hand If a users creates a context before loading symbols,  we call some gl:: functions to create the FBO, texture, etc, and we need symbols loaded for that. 

Some possible solutions I'm not in love with, but will work:
- Create a dummy context in NativeGLContext::getProcAdress if there isn't one bound
- Separate the context creation and the FBO initialization. The caller would have to load the symbols after the context creation and before the FBO initialization.

src/platform/with_wgl/native_gl_context.rs
MortimerGoro: It seems strange but we have to load and mix symbols from both because wglGetProcAddress​ doesn't return function pointers for some legacy functions that are exported by the opengl32.dll itself (old methods coming from OpenGL 1.1). I'll add some comments to the code to explain this.

src/platform/with_wgl/native_gl_context.rs
MortimerGoro: Another solution is to change the gl:: generator from global to struct based on this project. It requires a bit more work but it adds a lot of flexibility specially on windows were we can have multiple OpenGL implementations at the same time (for example angle and wgl). I want to add angle implementation for windows too and I had problems in servo because of the global gl. Struct based gl could be a good solution. What do you think?

emilio: I see...

In general I'd prefer to do the GL context dance on `get_proc_address`, to align with how the rest of back-ends behave. I think this shouldn't be a problem, and even though it certainly feels hacky, the other options do seem hacky too..

emilio: (Of course, the "have libraries locally per context" is a better option, but it's a lot more work, so probably can be tracked aside from this).

src/platform/with_wgl/native_gl_context.rs
```diff
@@ -0,0 +1,154 @@
...
+            })
+        } else {
+            None
+        }
+    }
+
+    fn current_handle() -> Option<Self::Handle> {
+        let handle = unsafe { wgl::GetCurrentContext() };
+        if !handle.is_null() {
+            let hdc = unsafe { wgl::GetCurrentDC() };
```
emilio: Should we `assert!` here that `hdc` is not null (or return an error)?

src/platform/with_wgl/utils.rs
```diff
@@ -0,0 +1,703 @@
...
+    let ctx = wgl::CreateContext(hdc as *const c_void);
+    if ctx.is_null() {
+        return Err(format!("wglCreateContext failed: {}",
+                           format!("{}", io::Error::last_os_error())));
+    }
+
+    if !share.is_null() {
+        if wgl::ShareLists(share as *const c_void, ctx) == 0 {
+            return Err(format!("wglShareLists failed: {}",
+                               format!("{}", io::Error::last_os_error())));
```
emilio: nit: this second `format!` (here and above) shouldn't be needed.

src/platform/with_wgl/utils.rs
```diff
@@ -0,0 +1,703 @@
...
+        }
+    };
+
+    return Ok((ctx as winapi::HGLRC, hdc));
```
emilio: nit: no need the `return` keyword at the end of the function, you can avoid the semicolon at the end of the line.

src/platform/with_wgl/utils.rs
```diff
@@ -0,0 +1,703 @@
...
+}
+
+// creates a full context: attempts to use optional ext WGL functions
+unsafe fn create_full_context(settings: &WGLAttributes,
```
emilio, line 88: This seems sensible enough, though, I'll need to study a bit more since I'm not confident enough with the windows API. Thanks for pointing to glutin.

Also, can you add proper attribution to the `glutin` code?

emilio (commented):
I need a bit more study still, but this looks great! Let me know if I can help somehow :)

src/platform/with_wgl/native_gl_context.rs
MortimerGoro: PR updated with the dummy context implementation ;)

src/platform/with_wgl/utils.rs
MortimerGoro: I have added more attribution to glutin code and link to their repo in the last commit. Maybe you could create a licence/author files in this repo and we can add more attribution there

src/platform/with_wgl/native_gl_context.rs
```diff
@@ -0,0 +1,181 @@
...
+            } else {
+                Err("wgl::makeCurrent failed")
+            }
+        }
+    }
+
+    fn unbind(&self) -> Result<(), &'static str> {
+        unsafe {
+            if self.is_current() && wgl::MakeCurrent(ptr::null_mut(), ptr::null_mut()) == 0 {
+                Err("gwl::MakeCurrent (on unbind)")
```
emilio: nit: `wgl` instead of `gwl`.

src/platform/with_wgl/utils.rs
```diff
@@ -0,0 +1,709 @@
...
+                                            attributes.as_ptr());
+
+    if ctx.is_null() {
+        return Err(format!("wglCreateContextAttribsARB failed: {}",
+                           io::Error::last_os_error()));
+    }
+
+    // Disable or enable vsync
+    if extensions.split(' ').find(|&i| i == "WGL_EXT_swap_control").is_some() {
+        let _guard = try!(CurrentContextGuard::make_current(hdc, ctx as winapi::HGLRC));
```
emilio: Are you sure you need the `CurrentContextGuard` here? It seems to me we're going to bind the context right away after this, so we can rather bind it directly.

src/platform/with_wgl/utils.rs
```diff
@@ -0,0 +1,709 @@
...
+        if extra.SwapIntervalEXT(if settings.vsync { 1 } else { 0 }) == 0 {
+            return Err("wglSwapIntervalEXT failed".to_owned());
+        }
+    }
+
+    Ok((ctx as winapi::HGLRC, hdc))
+}
+
+unsafe fn create_hidden_window() -> Result<winapi::HWND, &'static str> {
+
```
emilio: nit: Remove this extra line.

emilio (approved):
The code looks reasonable, I can't test it, but I'm happy with it. Please rebase, do a squash and then I'll merge it.

If you find a way to run the tests this using AppVeyor or similar, that'd be awesome :-)

src/platform/with_wgl/native_gl_context.rs
```diff
@@ -0,0 +1,179 @@
...
+unsafe impl Sync for NativeGLContextHandle {}
+
+impl NativeGLContextMethods for NativeGLContext {
+    type Handle = NativeGLContextHandle;
+
+    fn get_proc_address(addr: &str) -> *const () {
+        let addr = CString::new(addr.as_bytes()).unwrap();
+        let addr = addr.as_ptr();
+        unsafe {
+
```
emilio: nit: extra line.

src/platform/with_wgl/native_gl_context.rs
```diff
@@ -0,0 +1,179 @@
...
+            // wglGetProcAddress​ doesn't return function pointers for some legacy functions,
+            // (the ones coming from OpenGL 1.1)
+            // These functions are exported by the opengl32.dll itself,
+            // so we have to fallback to kernel32 getProcAddress if wglGetProcAddress​ return null
+            match *GL_LIB {
+                Some(ref lib) => kernel32::GetProcAddress(lib.0, addr) as *const _,
+                None => ptr::null_mut(),
+            }
+        }
+
```
emilio: nit: Here too

src/platform/with_wgl/utils.rs
```diff
@@ -0,0 +1,711 @@
...
+                             winapi::WS_POPUP | winapi::WS_CLIPSIBLINGS | winapi::WS_CLIPCHILDREN);
+
+    // creating a dummy invisible window
+    let dummy_window = {
+        // getting the rect of the real window
+        let rect = {
+            let mut placement: winapi::WINDOWPLACEMENT = mem::zeroed();
+            placement.length = mem::size_of::<winapi::WINDOWPLACEMENT>() as winapi::UINT;
+            if user32::GetWindowPlacement(window, &mut placement) == 0 {
+                panic!();
```
emilio: please panic with a descriptive message, or return an error.

src/platform/with_wgl/utils.rs
```diff
@@ -0,0 +1,711 @@
...
+                                          ptr::null_mut());
+
+        if win.is_null() {
+            return Err(format!("CreateWindowEx function failed: {}",
+                               io::Error::last_os_error()));
+        }
+
+        let hdc = user32::GetDC(win);
+        if hdc.is_null() {
+            let err = Err(format!("GetDC function failed: {}", io::Error::last_os_error()));
```
emilio: nit: `return Err(..)`. Also, it seems you need to clean up the window you just created.

src/platform/with_wgl/utils.rs
```diff
@@ -0,0 +1,711 @@
...
+                                          ptr::null_mut(),
+                                          kernel32::GetModuleHandleW(ptr::null()),
+                                          ptr::null_mut());
+
+        if win.is_null() {
+            return Err(format!("CreateWindowEx function failed: {}",
+                               io::Error::last_os_error()));
+        }
+
+        let hdc = user32::GetDC(win);
```
emilio, line 636: I assume `hdc` doesn't need any cleanup, am I right?

src/platform/with_wgl/utils.rs
```diff
@@ -0,0 +1,711 @@
...
+// We'd like to credit and thank all of the Glutin contributors for their work.
+// (https://github.com/tomaka/glutin)
+
+pub unsafe fn create_offscreen(shared_with: winapi::HGLRC,
+                               settings: &WGLAttributes)
+                               -> Result<(winapi::HGLRC, winapi::HDC), String> {
+
+    let window: winapi::HWND = try!(create_hidden_window());
+    let hdc = user32::GetDC(window);
+    if hdc.is_null() {
```
emilio: Also need to cleanup the window here.

src/platform/with_wgl/utils.rs
```diff
@@ -0,0 +1,711 @@
...
+        return Err("GetDC function failed".to_owned());
+    }
+
+    let extra = try!(load_extra_functions(window));
```
emilio: And here.

src/platform/with_wgl/utils.rs
```diff
@@ -0,0 +1,711 @@
...
+        let data = CStr::from_ptr(data).to_bytes().to_vec();
+        String::from_utf8(data).unwrap()
+
+    } else {
+        String::new()
+    };
+
+
+    let (id, _) = if extensions.split(' ').find(|&i| i == "WGL_ARB_pixel_format").is_some() {
+        try!(choose_arb_pixel_format(&extra, &extensions, hdc, &settings.pixel_format)
```
emilio: And here, etc...

src/platform/with_wgl/utils.rs
```diff
@@ -0,0 +1,711 @@
...
+        return Err("GetDC function failed".to_owned());
+    }
+
+    let extra = try!(load_extra_functions(window));
+
+    let extensions = if extra.GetExtensionsStringARB.is_loaded() {
+        let data = extra.GetExtensionsStringARB(hdc as *const _);
+        let data = CStr::from_ptr(data).to_bytes().to_vec();
+        String::from_utf8(data).unwrap()
+
```
emilio: nit: Spurious line

src/platform/with_wgl/utils.rs
```diff
@@ -0,0 +1,711 @@
...
+    } else if extra.GetExtensionsStringEXT.is_loaded() {
+        let data = extra.GetExtensionsStringEXT();
+        let data = CStr::from_ptr(data).to_bytes().to_vec();
+        String::from_utf8(data).unwrap()
+
```
emilio: nit: Spurious line.

src/platform/with_wgl/utils.rs
```diff
@@ -0,0 +1,711 @@
...
+    let extra = try!(load_extra_functions(window));
+
+    let extensions = if extra.GetExtensionsStringARB.is_loaded() {
+        let data = extra.GetExtensionsStringARB(hdc as *const _);
+        let data = CStr::from_ptr(data).to_bytes().to_vec();
+        String::from_utf8(data).unwrap()
+
+    } else if extra.GetExtensionsStringEXT.is_loaded() {
+        let data = extra.GetExtensionsStringEXT();
+        let data = CStr::from_ptr(data).to_bytes().to_vec();
```
emilio, line 43: nit: `into_vec` prevents an allocation here.

src/platform/with_wgl/utils.rs
```diff
@@ -0,0 +1,711 @@
...
+        String::new()
+    };
+
+
+    let (id, _) = if extensions.split(' ').find(|&i| i == "WGL_ARB_pixel_format").is_some() {
+        try!(choose_arb_pixel_format(&extra, &extensions, hdc, &settings.pixel_format)
+            .map_err(|_| "Pixel format not available".to_owned()))
+    } else {
+        try!(choose_native_pixel_format(hdc, &settings.pixel_format)
+            .map_err(|_| "Pixel format not available".to_owned()))
```
emilio: Would be nice using two different errors here.

src/platform/with_wgl/utils.rs
```diff
@@ -0,0 +1,711 @@
...
+                                     -> Result<(c_int, PixelFormat), ()> {
+    // TODO: hardware acceleration is not handled
+
+    // handling non-supported stuff
+    if reqs.float_color_buffer {
+        return Err(());
+    }
+
+    match reqs.multisampling {
+        Some(0) => (),
```
emilio: nit: `Some(0) | None => ()`, instead of two braces.

src/platform/with_wgl/utils.rs
```diff
@@ -0,0 +1,711 @@
...
+
+    // now querying
+    let pf_id = gdi32::ChoosePixelFormat(hdc, &descriptor);
+    if pf_id == 0 {
+        return Err(());
+    }
+
+    // querying back the capabilities of what windows told us
+    let mut output: winapi::PIXELFORMATDESCRIPTOR = mem::zeroed();
+    if gdi32::DescribePixelFormat(hdc,
```
emilio, line 516: does `pf_id` need cleanup?

src/platform/with_wgl/wgl_attributes.rs
```diff
@@ -0,0 +1,83 @@
+// Attributes to use when creating an OpenGL context.
+#[derive(Clone, Debug)]
+pub struct WGLAttributes {
+    //
```
emilio: nit: Empty comment, please remove.

src/platform/with_wgl/utils.rs
```diff
@@ -0,0 +1,711 @@
...
+unsafe fn create_basic_context(hdc: winapi::HDC,
+                               share: winapi::HGLRC)
+                               -> Result<(winapi::HGLRC, winapi::HDC), String> {
+    let ctx = wgl::CreateContext(hdc as *const c_void);
+    if ctx.is_null() {
+        return Err(format!("wglCreateContext failed: {}", io::Error::last_os_error()));
+    }
+
+    if !share.is_null() {
+        if wgl::ShareLists(share as *const c_void, ctx) == 0 {
```
emilio: You need to clean up the context if this fails.

emilio (changes requested):
Sorry I hadn't look more carefully the error paths, I've left a few comments on that to prevent leaks.

src/platform/with_wgl/utils.rs
MortimerGoro: can you give an example? I don't find good doc about into_vec

src/platform/with_wgl/utils.rs
MortimerGoro: No, its just a pixel format index and doesn't require a release call

MortimerGoro:
I have improved WGL & winapi disposal in error paths using a scoped struct. Also got rid of CurrentContextGuard code.

src/platform/with_wgl/utils.rs
```diff
@@ -596,35 +632,43 @@ unsafe fn load_extra_functions(window: winapi::HWND) -> Result<wgl_ext::Wgl, Str
...
-    Ok(wgl_ext::Wgl::load_with(|addr| {
+    let result = wgl_ext::Wgl::load_with(|addr| {
         let addr = CString::new(addr.as_bytes()).unwrap();
         let addr = addr.as_ptr();
         wgl::GetProcAddress(addr) as *const c_void
-    }))
+    });
+
+    // Unbind the current context so WGLScopedContext can release resources properly
+    // wgl::DeleteContex must be called with unbound contexts
```
emilio: nit: `DeleteContext`

src/platform/with_wgl/utils.rs
```diff
@@ -25,60 +25,61 @@ use super::wgl_ext;
...
     };
 
-    try!(set_pixel_format(hdc, id));
-
-    create_full_context(settings, &extra, &extensions, hdc, shared_with)
+    try!(set_pixel_format(ctx.device_ctx, id));
 
+    let result = create_full_context(settings, &extra, &extensions, ctx.device_ctx, shared_with);
+    if result.is_ok() {
+        ctx.dispose = false; // Everything is ok, don't dispose the scoped ctx
```
emilio: probably something like removing the `dispose` flag, and use `mem::forget` is better here? It's probably fine anyway.

src/platform/with_wgl/native_gl_context.rs
```diff
@@ -54,10 +54,13 @@ impl Drop for NativeGLContext {
     fn drop(&mut self) {
         unsafe {
             if !self.weak {
+                self.unbind().unwrap(); //wglDeleteContext requires unbind
```
emilio: Does this mean that the context to be deleted needs to be unbound? Or that there must be no current context.

I guess this is the first one, but please be clearer in this regard in the comment. Also, space after `//`.

src/platform/with_wgl/native_gl_context.rs
```diff
@@ -54,10 +54,13 @@ impl Drop for NativeGLContext {
...
                 wgl::DeleteContext(self.render_ctx as *const _);
                 let window = user32::WindowFromDC(self.device_ctx);
-                user32::ReleaseDC(window, self.device_ctx);
-                user32::DestroyWindow(window);
+                if !window.is_null() {
```
emilio: what has suddenly making `window` nullable? If it happened before and is something you've fixed, I guess that's fine.

But if it's because we can create multiple contexts associated with the same device context, and then have the window deleted, that can mean multiple things if I'm correct:
- The context doesn't require the window to stay alive, in which case we can delete it earlier.
- The context does require the window to stay alive, in which case this is wrong.
- The context does require the window to stay alive at least until we do sharing? That doesn't make sense to me but could be it.

Can you tell me which one is the relevant one?

src/platform/with_wgl/native_gl_context.rs
MortimerGoro: Improved comments

src/platform/with_wgl/utils.rs
MortimerGoro: I like the mem::forget approach, it seems more rusty

src/platform/with_wgl/native_gl_context.rs
MortimerGoro: We are not creating multiple contexts associated with the same HDC. It was just a extra safety check because null is a possible return value of WindowFromDC. It shouldn't happen in our code, but it seems safer to check for null.

emilio: Can we instead assert that it's not null, and handle it only if under some circumstances the assertion fails?

MortimerGoro: Done

emilio (approved):
Ok, let's do it. I don't think I can say too much right now with my limited wgl/winapi experience, and this has already stayed here for really long.

Sorry it took so long @MortimerGoro! Hopefully I wasn't too nitpicky :)

Thanks for this work!

If you look into how to test this (with AppVeyor, for example), so it doesn't regress, it'd be really nice :)

PR #65, opened 2016-09-20
Osmesa Backend

emilio:


emilio:
Sorry for not finishing this, @glennw, I have a OSMesa back-end integration WIP somewhere, but I couldn't finish it today (to many other compromises :/).

I just didn't want you to do this work in duplicate. Also, note that choosing a non-core backend is intentional, in my tests sharing works fine with these, but we can add the necessary code to WebGL to handle core profiles properly.

r? @glennw

emilio:
Note that I need to add the osmesa packages or something so our CI passes again, but that's not relevant to the core of this PR.

glennw:
@emilio Looks good to me. I think the CI issue is that travis uses Ubuntu 14.04 which has an old version of OSMesa that doesn't include the symbols we use.

I have a source package of OSMesa that builds via cargo here https://github.com/glennw/osmesa-src. We could add that as a dependency and set the env vars to the location of that to allow testing.

We could land it with the OSMesa test disabled for now, and add the source build of OSMesa above later, perhaps?

glennw:
I've been testing this locally today and it's working well.

r=me once the version number is bumped and we decide what to do about the linux + travis test issue.

emilio:
I made a quick try to enable CI and ran into linker errors for an old LLVM version. Opened #66 to track it.

emilio:
Merging this since this goes untested anyway.

PR #68, opened 2016-09-26
Implement multithreading tests

MortimerGoro:
I have added some multithreading tests. It happened to me that context sharing was working ok in the current test (same thread) but was failing on servo due to multithreading environment. Better to test all the scenarios ;)

### DIFF
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -12,6 +12,8 @@ use NativeGLContext;
 use NativeGLContextMethods;
 use GLContextAttributes;
 use ColorAttachmentType;
+use std::{thread, time};
+use std::sync::mpsc;
 
 #[cfg(target_os="macos")]
 #[link(name="OpenGL", kind="framework")]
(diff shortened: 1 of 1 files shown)


src/tests.rs
```diff
@@ -158,6 +160,119 @@ fn test_sharing() {
...
+        gl::clear(gl::COLOR_BUFFER_BIT);
+
+        let vec = gl::read_pixels(0, 0, size.width, size.height, gl::RGBA, gl::UNSIGNED_BYTE);
+        test_pixels_eq(&vec, &[0, 255, 0, 255]);
+
+        sender.send(()).unwrap();
+
+        // Avoid drop until test ends
+        loop {
+            thread::sleep(time::Duration::from_millis(200));
```
emilio: This is nasty ;-)

Please add another kind of mechanism to block, another pair of channels would work.

emilio:
Neat, looks good to me with that test change.

src/tests.rs
MortimerGoro: Agree ;) Already fixed

emilio:
r=me

PR #69, opened 2016-09-26
Add GLDispatcher traits to allow send GL commands to the thread of the shared context

MortimerGoro:
To make the WGL implementation fully compatible with shared GL contexts in multiple threads (the case of servo & webrender ;)) we need to dispatch some gl commands to a specific thread. The wglShareLists function fails if the context to share is current in a different thread. Additionally wglMakeCurrent cannot 'steal' a context that is current in other thread, so we have to unbind the shared context in its own thread, call wglShareLists in the thread of the new context and bind the original share context again when the wglShareList is completed.

Currently when we receive a shared handle we don't know in which thread is bound and running. Even if we knew we couldn't dispatch commands to that thread because the users of the library are responsible for handling the thread event queues and we want to keep this library generic and atomic.

To fix this problem I have added a GLDispatcher traits type. Now the caller can (optionally) include a dispatcher when creating a NativeContext. The API is still backwards compatbile because the previous contructors still work.

I have already tested some webgl animations in WGL running on servo with webrender enabled. The context sharing is working ok now with this approach.

I'll push the updated WGL branch and a PR to webrender/servo when this is merged.

### DIFF
--- a/src/gl_context.rs
+++ b/src/gl_context.rs
@@ -29,7 +29,13 @@ impl<Native> GLContext<Native>
     where Native: NativeGLContextMethods
 {
     pub fn create(shared_with: Option<&Native::Handle>) -> Result<GLContext<Native>, &'static str> {
-        let native_context = try!(Native::create_shared(shared_with));
+        Self::create_shared_with_dispatcher(shared_with, None)
+    }
+
+    pub fn create_shared_with_dispatcher(shared_with: Option<&Native::Handle>,
+                                         dispatcher: Option<Box<GLContextDispatcher>>)
+        -> Result<GLContext<Native>, &'static str> {
+        let native_context = try!(Native::create_shared_with_dispatcher(shared_with, dispatcher));
         try!(native_context.make_current());
         let attributes = GLContextAttributes::any();
         let formats = GLFormats::detect(&attributes);
@@ -44,7 +50,6 @@ impl<Native> GLContext<Native>
             limits: limits,
         })
     }
-
 
     #[inline(always)]
     pub fn get_proc_address(addr: &str) -> *const () {
@@ -61,9 +66,18 @@ impl<Native> GLContext<Native>
                color_attachment_type: ColorAttachmentType,
                shared_with: Option<&Native::Handle>)
         -> Result<GLContext<Native>, &'static str> {
+        Self::new_shared_with_dispatcher(size, attributes, color_attachment_type, shared_with, None)
+    }
+
+    pub fn new_shared_with_dispatcher(size: Size2D<i32>,
+                                      attributes: GLContextAttributes,
+                                      color_attachment_type: ColorAttachmentType,
+                                      shared_with: Option<&Native::Handle>,
+                                      dispatcher: Option<Box<GLContextDispatcher>>)
+        -> Result<GLContext<Native>, &'static str> {
         // We create a headless context with a dummy size, we're painting to the
         // draw_buffer's framebuffer anyways.
-        let mut context = try!(Self::create(shared_with));
+        let mut context = try!(Self::create_shared_with_dispatcher(shared_with, dispatcher));
 
         context.formats = GLFormats::detect(&attributes);
         context.attributes = attributes;
@@ -151,6 +165,12 @@ impl<Native> GLContext<Native>
     }
 }
 
+// Dispatches functions to the thread where a NativeGLContext is bound.
+// Right now it's used in the WGL implementation to dispatch functions to the thread
+// where the context we share from is bound. See the WGL implementation for more details.
+pub trait GLContextDispatcher {
+    fn dispatch(&self, Box<Fn() + Send>);
+}
 
 trait GLContextPrivateMethods {
     fn init_offscreen(&mut self, Size2D<i32>, ColorAttachmentType) -> Result<(), &'static str>;
--- a/src/platform/with_wgl/native_gl_context.rs
+++ b/src/platform/with_wgl/native_gl_context.rs
@@ -1,7 +1,9 @@
 use platform::NativeGLContextMethods;
+use gl_context::GLContextDispatcher;
 use std::ffi::CString;
 use std::os::raw::c_void;
 use std::ptr;
+use std::sync::mpsc;
 
 use winapi;
 use user32;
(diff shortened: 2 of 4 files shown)


MortimerGoro:
We can also use this feature in any API.  For example in the case we need to query parameters of the shared context without making it current in the current thread. If that context was already bound to another thread, it can lead to problems. It's unsafe to have the same context bound to 2 different threats at the same time.

MortimerGoro:
You can check the changes to webrender, servo and WGL in these branches:

https://github.com/MortimerGoro/rust-offscreen-rendering-context/commit/a770d95034e4a637915d644387140e6071be4e9a
https://github.com/MortimerGoro/servo/commit/64f92be79ad5cc2df2fd207a5d7fc085fbb1c6a9
https://github.com/MortimerGoro/webrender/commit/b9a601c802d516dd5804ea8bcbbf1e311df631a4

emilio:
Hmm... So I'm not too fond on this approach.

I'm not sure if I've understood the problem though, can you help me confirm?

On one hand, we have the webrender API thread, where all the WebGL contexts live. On the other hand, we have Servo's main thread, where Webrender's main context lives, which is shared with the others. No context is bound to multiple threads at the same time.

Is the fact that we have two contexts sharing stuff bound to different contexts a problem in `wgl`? What kind of problems does it generate?

I can imagine that the problem may be we're sending draw calls to both contexts at the same time? In that case, I guess it's better to make the `render` function in WR notify the back end when we've ended the update, and make the back-end buffer messages from WebGL until that happens? It seems like a more self-contained solution.

MortimerGoro:
> Hmm... So I'm not too fond on this approach.
> 
> I'm not sure if I've understood the problem though, can you help me confirm?
> 
> On one hand, we have the webrender API thread, where all the WebGL contexts live. On the other hand, we have Servo's main thread, where Webrender's main context lives, which is shared with the others. No context is bound to multiple threads at the same time.

Yes, I understood the same scenario checking the code and logging current threads. We have the webrender thread where all WebGL contexts live and run gl commands. The webrender's main context uses Servo's main context & main thread, and all ctxs are bound only to a single thread.

> Is the fact that we have two contexts sharing stuff bound to different contexts a problem in wgl? What kind of problems does it generate?

No, that't not a problem in WGL. The mutithreading testcase works and in my local branch I can run a rotating cube webgl demo with webrender enabled, even with dom elements in the same page. So, webrender's ctx and webgl ctx are running fine on different WGL contexts bound to different threads at the same time.

> I can imagine that the problem may be we're sending draw calls to both contexts at the same time? 

The problem happens in the exact moment of the shared context creation, when we call NativeGLContext::new(...) (shared_with param = webderer's main context handle). 

In WGL to create a shared context we need to call wglShareLists function. But this function has some special requirements. The context where we want to share from cannot be bound to any thread (just during the function call, it doesn't matter before or after). In our scenario, to make wglShareLists not fail, the context that can't be bound to any thread is webrender's main ctx thread == servo's main thread.

So, in order to wglShareLists call succeed, before calling the function we have to unbind the WR main ctx. We have to do that in its own thread, because we cant 'steal' a context that is current in other thread. Once the context is unbound, we call wglShareLists (in the WebGL thread) and it succeeds. After that we restore the previous state of the WR's thread, binding it again (in the WR therad).

> In that case, I guess it's better to make the render function in WR notify the back end when we've ended the update, and make the back-end buffer messages from WebGL until that happens? It seems like a more self-contained solution.

IMO, the GLDispatcher attached to the shared_with_handle in the NativeGLContext API is a more generic solution. Some reasons:
- It allows any combination of shared contexts, like separated lists of shared contexts running in any thread. I think that this library can be used among many different projects and each one might have its own scenario.
- In WR we have this scenario now, but will it be the same when we have multiple windows and multiple tabs, or if we implement webgl sharing in webworkers?
- Can you guarantee that render function is called every frame? I mean that render function might not be called if there are dirty checks in servo's compositor and the view's don't need to be updated. 
- It only dispatches functions when needed
- While the context is created calling the NativeGLContext::new() func, WR doesn't have the instance yet. In order to notify in the render func while the instance is being created,  we'd also need an extra argument in the NativeGLContext creation like the one I have added.

The GLDispatcher trait can also be implemented using the WR's render function to run pending dispathed functions. If you can guarantee that it is called every frame or that the dirty checks in Servo's compositor are and not will be a problem,  the PR to servo can be avoided.  I'm sure that servo's main event queue runs everything but not sure about the current and future WR.

emilio:
Ok, since these are reasonable arguments, I guess in this case I'm fine with this, as long as it's really well documented in which case and for which backends we use this, and also which constraints are imposed on the dispatcher for sharing in wgl to work properly (basically, that the dispatcher if present must dispatch to the thread where the pre-existing context to share context is bound).

Ideally:
- We'll find a better name for this. `GLDispatcher` is a really generic name.
- We wouldn't require a new type that only on some circumstances predates the `handle`. Can we add instead an `Option<Box<GLDispatcher>>` param?

MortimerGoro:
> We wouldn't require a new type that only on some circumstances predates the handle. Can we add instead an Option<Box<GLDispatcher>> param?

Ok, I have improved the code to add the dispatcher using a parameter. Also added a default implementation in the NativeGLContextMethods traits so we don't have to change any code in the APIs that don't use the dispatcher. 

P.D: Coming from C++, I really miss function overloading in Rust. It would be very useful in this case to add the new parameter without adding more function names.

> We'll find a better name for this. GLDispatcher is a really generic name.

I changed it to GLContextDispatcher. If you have thought a better name tell me and I'll update it ;)

src/gl_context.rs
```diff
@@ -29,7 +29,13 @@ impl<Native> GLContext<Native>
     where Native: NativeGLContextMethods
 {
     pub fn create(shared_with: Option<&Native::Handle>) -> Result<GLContext<Native>, &'static str> {
-        let native_context = try!(Native::create_shared(shared_with));
+        Self::create_shared(shared_with, None)
+    }
+
+    pub fn create_shared(shared_with: Option<&Native::Handle>,
```
emilio: Maybe `shared_with_dispatcher` (or `create_shared_with_dispatcher`)? `create_shared` is a bit confusing, since the normal `create` function also creates a shared context.

src/gl_context.rs
```diff
@@ -61,9 +66,18 @@ impl<Native> GLContext<Native>
                color_attachment_type: ColorAttachmentType,
                shared_with: Option<&Native::Handle>)
         -> Result<GLContext<Native>, &'static str> {
+        Self::new_shared(size, attributes, color_attachment_type, shared_with, None)
+    }
+
+    pub fn new_shared(size: Size2D<i32>,
+               attributes: GLContextAttributes,
```
emilio: nit: Align the parameters to match the alignment of `size`.

src/platform/with_wgl/native_gl_context.rs
```diff
@@ -108,17 +110,58 @@ impl NativeGLContextMethods for NativeGLContext {
...
+                    let handle = NativeGLContextHandle(render_ctx, device_ctx);
+                    dispatcher.dispatch(Box::new(move || {
+                        unsafe { 
+                            if wgl::MakeCurrent(handle.1 as *const _, handle.0 as *const _) == 0 {
+                                error!("WGL MakeCurrent failed!");
+                            } 
+                        };
+                        tx.send(()).unwrap();
+                    }));
+                    // Wait uintil the wglMakeCurrent operation is completed in the thread of the shared context
```
emilio: nit: `until`.

emilio (approved):
> I changed it to GLContextDispatcher. If you have thought a better name tell me and I'll update it ;)

Fair enough, can't think of a better name right now :)

Just a few nits, then r=me. The patch looks way cleaner now :)

emilio:
Ok, there we go. Thanks for the patch @MortimerGoro!

emilio:
All your patches are published now as 0.4.5 Imanol :)

PR #71, opened 2016-10-05
Fix CGL multithreading error

MortimerGoro:
Multithreading tests fail with Err("CGLChoosePixelFormat") error  on a MacBook with a clean macOS sierra install. 

It seems that CGLChoosePixelFormat can fail if multiple threads try to open a display connection simultaneously. Fixed the issue adding a static mutex guard.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,4 +38,6 @@ winapi = "0.2"
 gdi32-sys = "0.2"
 user32-sys = "0.2"
 kernel32-sys = "0.2"
+
+[target.'cfg(any(target_os="macos", target_os="windows"))'.dependencies]
 lazy_static = "0.2"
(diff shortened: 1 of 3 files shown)


emilio:
Well, this is probably ok. Please use a [`StaticMutex`](https://doc.rust-lang.org/std/sync/struct.StaticMutex.html).

MortimerGoro:
> Please use a StaticMutex.

It seems to have been removed from the std API, or do I need to use a nightly rust compiler? I'm testing with the latest stable 1.12.0

emilio:
Hm... You're right, sorry. Can you use `.lock().unwrap()` instead of just `.lock()`, handle the error, or mention that you're explicitly ignoring it?

MortimerGoro:
Done!

emilio:
Thanks!

PR #73, opened 2016-10-20
Improve WGL shared context creation error paths

MortimerGoro:
The result of the first wglMakeCurrent call is handled through the channel. 

Also found a new error path. If the context creation itself fails we have always to restore the shared context's wglMakeCurrent values.

If the last wglMakeCurrent fails (very unlikely) it's not really a GLContext initialization problem, but we might cause some dirty status in the main thread . Currently it logs an error. Maybe panic?


emilio:
Looks great, thanks!

PR #74, opened 2016-11-17
Drop EGL depedency unless needed

staktrace:
Right now when building webrender on Linux (with Gecko), this crate pulls in a libEGL dependency which is really not needed, as it's only required as part of a testing feature. This commit fixes that.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "offscreen_gl_context"
 license = "MIT / Apache-2.0"
-version = "0.4.5"
+version = "0.5.0"
 authors = ["Emilio Cobos Álvarez <ecoal95@gmail.com>", "The Servo Project Developers"]
 description = "Creation and manipulation of HW accelerated offscreen rendering contexts in multiple platforms. Originally intended for the Servo project's WebGL implementation."
 repository = "https://github.com/emilio/rust-offscreen-rendering-context"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "offscreen_gl_context"
 license = "MIT / Apache-2.0"
-version = "0.4.5"
+version = "0.4.6"
```
emilio: Can you bump to 5.0 instead? This is technically a breaking change.

emilio (approved):
So the idea for this was that eventually we'd have different back-end support inside Linux, so you could use the default `NativeGLContext` (glx), or the egl variant.

There was a project to make servo run with EGL, which I guess never ended up making it, so I suppose this is fine.

Please make the version bump as breaking, and I'll merge this, thanks!

Cargo.toml
emilio: (0.5.0 i mean)

staktrace:
Thanks. Could you also please push the updated version to crates.io? In the firefox build we can't (for now, anyway) pull in repos from git, it has to be on crates.io.

emilio:
Actually I think you also need to modify `build.rs` so it doesn't print the link libEGL line, sorry for not noticing before.

And sure, I'll publish this as soon as it's merged.

emilio:
Thanks!

staktrace:
Thank you for the quick turnaround!

PR #75, opened 2016-11-30
Clear draw buffer after creation and resize.

MortimerGoro:
I've been working on fixing the scissor test for https://github.com/servo/servo/pull/14075 and found that the GLContexts are not cleared after creation. That's why that test is failing now (after fixing what the spec says about scissor).

I'm adding the clear changes here because I think that clearing the created framebuffer is a good default for any project.  I had a lot of problems in the past with buggy Adreno GPUs on Android because of this. A initial clear is a safeguard to avoid problems. Some Adrenos were such bad that even required clearing the Depth/Stencil even if the buffer is neither created and used.

### DIFF
--- a/src/gl_context.rs
+++ b/src/gl_context.rs
@@ -158,7 +158,7 @@ impl<Native> GLContext<Native>
         if self.draw_buffer.is_some() {
             let color_attachment_type =
                 self.borrow_draw_buffer().unwrap().color_attachment_type();
-            self.create_draw_buffer(size, color_attachment_type)
+            self.init_offscreen(size, color_attachment_type)
         } else {
             Err("No DrawBuffer found")
         }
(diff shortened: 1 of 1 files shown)


emilio:
I don't think we can call `ClearColor` here, or at least we should restore the clear color on the WebGL implementation, right?

Otherwise we're changing the underlying state of the GL context under the hood, and I while I think `Viewport` and `Scissor` are probably fine to do on resize, `ClearColor` is more dubious to me.

What do you think?

MortimerGoro:
Yeah, that can be a problem in the resize method. It's not a problem in the init because a transparent color is the default one.

We can restore the ClearColor from WebGL (we are restoring scissor value too now to mach the Spec).

emilio:
Ok, r=me with that in Servo then.

MortimerGoro:
Ok, please bump version ;)

emilio:
Done :P.

Next time feel free to PR the version bump too.

PR #80, opened 2017-02-03
Update gleam version constraint

mbrubeck:
1be3970a5702fa4696fcc3c4a98ec3bf353e9953 depends on gleam 0.2.31 or newer.


emilio:
We should seriously hook bors here. I'm happy to move the repo if wanted/needed btw.

PR #81, opened 2017-02-13
Update gleam, and use a dynamic GL function table. 

emilio:
cc @sotaroikeda

### DIFF
--- a/src/draw_buffer.rs
+++ b/src/draw_buffer.rs
@@ -1,6 +1,7 @@
 use euclid::Size2D;
 use gleam::gl;
 use gleam::gl::types::{GLuint, GLenum, GLint};
+use std::rc::Rc;
 
 use GLContext;
 use NativeGLContextMethods;
@@ -18,10 +19,11 @@ impl Default for ColorAttachmentType {
 }
 
 
-/// We either have a color renderbuffer
-/// Or a surface bound to a texture
-/// bound to a framebuffer as a color
-/// attachment
+/// We either have a color renderbuffer, or a surface bound to a texture bound
+/// to a framebuffer as a color attachment.
+///
+/// NB: The draw buffer manages it, and calls its destroy method on drop, this
+/// is just to avoid propagating the GL functions pointer further down.
 #[derive(Debug)]
 pub enum ColorAttachment {
     Renderbuffer(GLuint),
@@ -35,13 +37,11 @@ impl ColorAttachment {
             ColorAttachment::Texture(_) => ColorAttachmentType::Texture,
         }
     }
-}
 
-impl Drop for ColorAttachment {
-    fn drop(&mut self) {
-        match *self {
-            ColorAttachment::Renderbuffer(id) => gl::delete_renderbuffers(&[id]),
-            ColorAttachment::Texture(tex_id) => gl::delete_textures(&[tex_id]),
+    fn destroy(self, gl: &gl::Gl) {
+        match self {
+            ColorAttachment::Renderbuffer(id) => gl.delete_renderbuffers(&[id]),
+            ColorAttachment::Texture(tex_id) => gl.delete_textures(&[tex_id]),
         }
     }
 }
@@ -52,6 +52,7 @@ impl Drop for ColorAttachment {
 /// a depth or stencil buffer, depending on context
 /// requirements.
 pub struct DrawBuffer {
+    gl_: Rc<gl::Gl>,
     size: Size2D<i32>,
     framebuffer: GLuint,
     stencil_renderbuffer: GLuint,
@@ -61,14 +62,15 @@ pub struct DrawBuffer {
 }
 
 /// Helper function to create a render buffer
-/// TODO(ecoal95): We'll need to switch between `glRenderbufferStorage` and
-///   `glRenderbufferStorageMultisample` when we support antialising
-fn create_renderbuffer(format: GLenum,
+/// TODO(emilio): We'll need to switch between `glRenderbufferStorage` and
+/// `glRenderbufferStorageMultisample` when we support antialising
+fn create_renderbuffer(gl_: &gl::Gl,
+                       format: GLenum,
                        size: &Size2D<i32>) -> GLuint {
-    let ret = gl::gen_renderbuffers(1)[0];
-    gl::bind_renderbuffer(gl::RENDERBUFFER, ret);
-    gl::renderbuffer_storage(gl::RENDERBUFFER, format, size.width, size.height);
-    gl::bind_renderbuffer(gl::RENDERBUFFER, 0);
+    let ret = gl_.gen_renderbuffers(1)[0];
+    gl_.bind_renderbuffer(gl::RENDERBUFFER, ret);
+    gl_.renderbuffer_storage(gl::RENDERBUFFER, format, size.width, size.height);
+    gl_.bind_renderbuffer(gl::RENDERBUFFER, 0);
 
     ret
 }
@@ -77,7 +79,7 @@ impl DrawBuffer {
     pub fn new<T: NativeGLContextMethods>(context: &GLContext<T>,
                                           mut size: Size2D<i32>,
                                           color_attachment_type: ColorAttachmentType)
-                                          -> Result<DrawBuffer, &'static str>
+                                          -> Result<Self, &'static str>
     {
         const MIN_DRAWING_BUFFER_SIZE: i32 = 16;
         use std::cmp;
@@ -101,6 +103,7 @@ impl DrawBuffer {
         size.height = cmp::max(MIN_DRAWING_BUFFER_SIZE, size.height);
 
         let mut draw_buffer = DrawBuffer {
+            gl_: context.clone_gl(),
             size: size,
             framebuffer: 0,
             color_attachment: None,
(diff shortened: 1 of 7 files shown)


emilio:
Blocked on https://github.com/servo/gleam/pull/106

kvark:
So `gl::Gl` is a trait, and thus the new code works with the trait object, right?
Is there any benefit of using the unsafe raw pointers versus just `Rc`/`Weak`?

emilio:
@kvark That's right. It avoids some refcount churn. In this case it seemed easy enough to do it, but I'd be fine also with Rc. It may make sense to share gl function tables across threads though, that's something that Rc would prevent, but we don't use that at all I think so it may be ok.

emilio:
Oh, and I hadn't been following the gleam PR btw, the generic proposal looks ok to me. It's kind of sad to use dynamic dispatch, but I was taking that for granted already. Couldn't we make `Gl` a static table and add convenience methods to it instead? I did something similar for `clang-sys` for bindgen (see the functions struct in https://github.com/KyleMayes/clang-sys/pull/45). It'd be a matter of adding all the glue that would be inlined and would just look like:

```rust
#[inline]
fn flush(&self) {
    (self.functions.expect("glFlush not loaded"))()
}
```

sotaroikeda:
>  Couldn't we make Gl a static table and add convenience methods to it instead?

It might be better for performance. But  I want to handle it as another issue, since we want to enable Graphics branch soon.

kvark:
@emilio 
> It avoids some refcount churn.

The extra refcount work would only apply to resource creation/deletion, right? I wouldn't be too concerned and preferred the code clarity (and less unsafety) instead.

> It may make sense to share gl function tables across threads though

Could you explain why we'd need that? It sounds like a hack.

> Couldn't we make Gl a static table and add convenience methods to it instead?

Possible, although I'd rather want to see static dispatch here in the first place. Considering the deadlines, this will have to wait.

emilio:
On Wed, Feb 15, 2017 at 09:43:07PM -0800, Dzmitry Malyshau wrote:
> @emilio 
> > It avoids some refcount churn.
> 
> The extra refcount work would only apply to resource creation/deletion, right? I wouldn't be too concerned and preferred the code clarity (and less unsafety) instead.

Right, and that's a fair point.

> > It may make sense to share gl function tables across threads though
> 
> Could you explain why we'd need that? It sounds like a hack.

Didn't have a specific use case in mind, but yeah, I guess context
creation time is not a big deal.

I could imagine us sharing the same function hooks or something, but
yeah, as I said I don't expect it to be super-useful.

> > Couldn't we make Gl a static table and add convenience methods to it instead?
> 
> Possible, although I'd rather want to see static dispatch here in the
> first place. Considering the deadlines, this will have to wait.

That's fair, as I said I don't love dynamic dispatch, but we can
probably live with it.

kvark:
@emilio thanks!
I've just published gleam-0.3.0, so let's have this PR updated.

sotaroikeda:
@emilio, thank you!

emilio:
Done, feel free to take a look, let me know when do you want to merge this.

src/gl_context.rs
```diff
@@ -26,22 +32,25 @@ pub struct GLContext<Native> {
...
         Self::create_shared_with_dispatcher(shared_with, None)
     }
 
     pub fn create_shared_with_dispatcher(shared_with: Option<&Native::Handle>,
                                          dispatcher: Option<Box<GLContextDispatcher>>)
-        -> Result<GLContext<Native>, &'static str> {
+        -> Result<Self, &'static str> {
         let native_context = try!(Native::create_shared_with_dispatcher(shared_with, dispatcher));
+        // XXX how to handle GLES instantiation?
+        let gl_ = gl::GlFns::load_with(|s| GLContext::<Native>::get_proc_address(s) as *const _);
```
kvark: it appears that we need to sort this out before merging

src/draw_buffer.rs
```diff
@@ -35,13 +37,11 @@ impl ColorAttachment {
...
         }
     }
-}
 
-impl Drop for ColorAttachment {
-    fn drop(&mut self) {
-        match *self {
-            ColorAttachment::Renderbuffer(id) => gl::delete_renderbuffers(&[id]),
-            ColorAttachment::Texture(tex_id) => gl::delete_textures(&[tex_id]),
+    fn destroy(self, gl: &gl::Gl) {
```
kvark, line 41: why do we need to `destroy` now instead of `drop`?

kvark (changes requested):

src/gl_context.rs
emilio: Oh, right, I just grabbed that from the original Sotaro's patch. I guess the idea here would be doing so per-backend (that is, the egl back-end would load gles functions, etc). That being said, in which case would it make a difference? I guess we can provide only the GLES-like interface given this is primarily for WebGL and we create compat contexts, alternatively, I can make it generic, but that cries to avoid the virtual dispatch then.

src/draw_buffer.rs
emilio: Didn't want to grow them by an extra pointer given we only create one per draw buffer, but I can definitely do it I guess.

kvark: Ah, you are passing `gl` in, I see. And the function is not public, so there is no concern about calling it twice.
Let's leave it with `destroy` then :)

src/gl_context.rs
```diff
@@ -65,16 +74,20 @@ impl<Native> GLContext<Native>
                attributes: GLContextAttributes,
                color_attachment_type: ColorAttachmentType,
                shared_with: Option<&Native::Handle>)
-        -> Result<GLContext<Native>, &'static str> {
-        Self::new_shared_with_dispatcher(size, attributes, color_attachment_type, shared_with, None)
+        -> Result<Self, &'static str> {
+        Self::new_shared_with_dispatcher(size,
```
kvark, line 81: I'm a bit confused about this. `create_shared_with_dispatcher` calls `new_shared_with_dispatcher`, and `new_shared_with_dispatcher` calls `create_shared_with_dispatcher`, both seem unconditional.

src/gl_context.rs
kvark: > alternatively, I can make it generic, but that cries to avoid the virtual dispatch then

Well, doing static dispatch is still on the table, just for a different time frame - https://github.com/servo/gleam/issues/107
So maybe it's not the worst idea to be ready for that ;)

Really though, If this newly created context is shared with some existing one, we should use Gl function loader if that existing one is GL, and GLES correspondingly. If it's an entirely new one (`shared_with.is_none() == true`), this is probably where we want the `gleam::GlType` flag to be passed in for run-time context type selection.

kvark (changes requested):

src/gl_context.rs
emilio: It... doesn't? `create_shared_with_dispatcher` calls the relevant method on the underlying platform-dependent context.

src/gl_context.rs
emilio: That sounds reasonable, though probably for `WebGL` it's enough with loading unconditionally just the gles functions.

emilio: tl;dr; I'll prepare a patch with that.

sotaroikeda:
What is a current status?

emilio:
Whoops, totally forgot about this, there it is.

src/draw_buffer.rs
```diff
@@ -113,8 +116,9 @@ impl DrawBuffer {
 
         try!(draw_buffer.init(context, color_attachment_type));
 
-        debug_assert!(gl::check_frame_buffer_status(gl::FRAMEBUFFER) == gl::FRAMEBUFFER_COMPLETE);
-        debug_assert!(gl::get_error() == gl::NO_ERROR);
+        debug_assert!(
```
kvark: `debug_assert_eq!` would be much better here since it would print the actual error/status

src/draw_buffer.rs
```diff
@@ -149,127 +153,130 @@ impl DrawBuffer {
...
-// parent with Rc<GLContext> and call make_current()
-impl Drop for DrawBuffer {
-    fn drop(&mut self) {
-        gl::delete_framebuffers(&[self.framebuffer]);
 
-        // NOTE: Color renderbuffer is destroyed on drop of
-        //   ColorAttachment
-        gl::delete_renderbuffers(&[self.stencil_renderbuffer, self.depth_renderbuffer]);
+    fn gl(&self) -> &gl::Gl {
+        // See the comment on top the GLContext field to see why our usage of
```
kvark: perhaps, there is no longer a point in this comment? There is no `unsafe` here, so safety is assumed by default

src/gl_context.rs
kvark: I see now, you are correct.

src/tests.rs
```diff
@@ -23,28 +22,21 @@ extern {}
 #[link(name="GL")]
 extern {}
 
-#[cfg(not(target_os="android"))]
-static LOAD_GL: Once = ONCE_INIT;
+#[cfg(target_os = "android")]
+const GL_API_TYPE: gl::GlType = gl::GlType::Gles;
```
kvark: could use `default()`, which is already implemented on `GlType`

kvark (approved):
Thanks @emilio , looks good!
I got a few minor things, which are fine to address separately, if you wish.

emilio:
Thanks for the review @kvark! Landed and published as 0.7.0

sotaroikeda:
Thank you!

PR #89, opened 2017-03-09
platform: Allow not depending on x11 and using only OSMesa.

### LINKED ISSUE #88: Add a headless compilation mode.
It would be great if we could build on headless machines that don't have X11 installed. Perhaps a headless feature where only the OSMesa support is enabled?

emilio:
Fixes #88 

You can build this with `--no-default-features --features osmesa`. I've tested it with a server without X11 installed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "offscreen_gl_context"
 license = "MIT / Apache-2.0"
-version = "0.8.0"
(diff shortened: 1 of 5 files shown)


glennw:
I've tested on mac and it builds fine with:

```cargo build``` and ```cargo build --no-default-features --features=osmesa```

emilio:
Nice then! :)

PR #94, opened 2017-03-30
Fix eglGetProcAddress for core OpenGL entrypoints

MortimerGoro:
Hi,

The upgrade to the new gleam has broken offscreen EGL based contexts & Android WebGL on Servo. This is because eglProcAddress can't be used to resolve core OpenGL function entry points on many implementations. Loading core functions with eglProcAddress is only possible in EGL 1.5.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "offscreen_gl_context"
 license = "MIT / Apache-2.0"
-version = "0.8.4"
+version = "0.8.5"
 authors = ["Emilio Cobos Álvarez <emilio@crisal.io>", "The Servo Project Developers"]
 description = "Creation and manipulation of HW accelerated offscreen rendering contexts in multiple platforms. Originally intended for the Servo project's WebGL implementation."
 repository = "https://github.com/emilio/rust-offscreen-rendering-context"
@@ -14,7 +14,7 @@ gl_generator = "0.5"
 default = ["x11"]
 osmesa = ["osmesa-sys"]
 # NOTE: Just for testing use, there are no other changes
-test_egl_in_linux = []
+test_egl_in_linux = ["libloading", "lazy_static"]
 
 [dependencies]
 log  = "0.3"
@@ -22,6 +22,8 @@ gleam = "0.4"
 euclid = "0.11"
 serde = { version = "0.9", optional = true }
 osmesa-sys = { version = "0.1", optional = true }
+libloading = { version = "0.3", optional = true }
+lazy_static = { version = "0.2", optional = true }
 
 [target.x86_64-apple-darwin.dependencies]
 core-foundation = "0.3.0"
@@ -38,5 +40,8 @@ gdi32-sys = "0.2"
 user32-sys = "0.2"
 kernel32-sys = "0.2"
 
-[target.'cfg(any(target_os="macos", target_os="windows"))'.dependencies]
+[target.'cfg(any(target_os="macos", target_os="windows", target_os="android"))'.dependencies]
 lazy_static = "0.2"
+
+[target.'cfg(target_os = "android")'.dependencies]
+libloading = "0.3"
--- a/src/platform/with_egl/native_gl_context.rs
+++ b/src/platform/with_egl/native_gl_context.rs
@@ -9,11 +9,14 @@ use libloading as lib;
 
 lazy_static! {
     static ref GL_LIB: Option<lib::Library>  = {
-        if cfg!(target_os = "android") {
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -38,5 +40,8 @@ gdi32-sys = "0.2"
 user32-sys = "0.2"
 kernel32-sys = "0.2"
 
-[target.'cfg(any(target_os="macos", target_os="windows"))'.dependencies]
+[target.'cfg(any(target_os="macos", target_os="windows", target_os="android"))'.dependencies]
 lazy_static = "0.2"
```
kvark, line 44: interesting, so we can have a dependency mentioned twice?

src/platform/with_egl/native_gl_context.rs
```diff
@@ -2,9 +2,20 @@ use euclid::Size2D;
...
 use std::ffi::CString;
+use std::ops::Deref;
 use egl;
 use egl::types::{EGLint, EGLBoolean, EGLDisplay, EGLSurface, EGLConfig, EGLContext};
+use libloading as lib;
 
+lazy_static! {
+    static ref GL_LIB: Option<lib::Library>  = {
+        if cfg!(target_os = "android") {
+            lib::Library::new("libGLESv2.so").ok()
```
kvark: This doesn't support run-time selection of the GL backend. E.g. one may want to run with GLES on Linux desktop.

MortimerGoro: That would require a change in the API because get_proc_address trait is static and we can't know the user preference. We could work on that on a separate issue/PR, it's less critical in this moment.

src/platform/with_egl/native_gl_context.rs
```diff
@@ -69,10 +80,21 @@ impl Drop for NativeGLContext {
...
+    // retrieve extension functions. Some implementatios return NULL for core OpenGL functions.
+    // Other implementations may return non-NULL values even for invalid core or extension symbols.
+    // This is very dangerous, so we use dlsym function before calling eglGetProcAddress
+    // in order to avoid possible garbage pointers.
     fn get_proc_address(addr: &str) -> *const () {
         unsafe {
-            let addr = CString::new(addr.as_bytes()).unwrap().as_ptr();
-            egl::GetProcAddress(addr as *const _) as *const ()
+            if let Some(ref lib) = *GL_LIB {
+                let symbol: lib::Symbol<unsafe extern fn()> = lib.get(addr.as_bytes()).unwrap();
```
emilio: Perhaps returning `ptr::null` instead of unwrapping?

src/platform/with_egl/native_gl_context.rs
```diff
@@ -69,10 +80,21 @@ impl Drop for NativeGLContext {
...
+                return *symbol.deref() as *const();
```
emilio: nit: space after `const`

src/platform/with_egl/native_gl_context.rs
emilio: Perhaps we should be using `or_else` and not platform-cfgs?

`Library::new("libGLESv2.so").or_else(|| lib::Library::new("libGL.so"))`?

MortimerGoro:
Curiosity: using or_else or other closuse patterns within the lazy_static macro causes some obscure errors: possibly related issue here https://github.com/rust-lang/rust/issues/24680

src/platform/with_egl/native_gl_context.rs
```diff
@@ -9,11 +9,15 @@ use libloading as lib;
 
 lazy_static! {
     static ref GL_LIB: Option<lib::Library>  = {
-        if cfg!(target_os = "android") {
-            lib::Library::new("libGLESv2.so").ok()
-        } else {
-            lib::Library::new("libGL.so").ok()
-        }
+       let names = ["libGLESv2.so", "libGLES.so", "libGLESv3.so"];
```
emilio: Still need `libGL`?

src/platform/with_egl/native_gl_context.rs
```diff
@@ -9,11 +9,15 @@ use libloading as lib;
...
+       for name in &names {
+           let lib = lib::Library::new(name);
+           if lib.is_ok() {
```
emilio:
```
if let Ok(lib) = lib::Library::new(name) {
    return Some(lib);
}
```

May be slightly clearer, your call.

MortimerGoro:
@emilio I added libGL.so for desktop GL on linux. Ideally we should have a way to choose the desired library version instead of the predefined order.

emilio:
LGTM, thanks :)

PR #95, opened 2017-04-03
Fix EGL compilation error

MortimerGoro:
I pushed a compilation error in the latest PR after compiling for desktop with test_egl_in_linux disabled... 

I'll add android compilation to travis to prevent this from happening again.


emilio:
LGTM

PR #97, opened 2017-04-10
Add support for packed depth stencil attachment.

MortimerGoro:
Add support for packed depth stencil attachment. Fix incomplete framebuffer error on some platforms when using both depth and stencil attachments.

### DIFF
--- a/src/gl_context.rs
+++ b/src/gl_context.rs
@@ -25,6 +25,7 @@ pub struct GLContext<Native> {
     capabilities: GLContextCapabilities,
     formats: GLFormats,
     limits: GLLimits,
+    extensions: Vec<String>
 }
 
 impl<Native> GLContext<Native>
@@ -47,8 +48,10 @@ impl<Native> GLContext<Native>
         };
 
         try!(native_context.make_current());
+        let extensions = gl_.get_string(gl::EXTENSIONS);
+        let extensions: Vec<String> = extensions.split(&[',',' '][..]).map(|s| s.into()).collect();
         let attributes = GLContextAttributes::any();
-        let formats = GLFormats::detect(&attributes, &*gl_);
+        let formats = GLFormats::detect(&attributes, &extensions[..]);
         let limits = GLLimits::detect(&*gl_);
 
         Ok(GLContext {
@@ -59,6 +62,7 @@ impl<Native> GLContext<Native>
             capabilities: GLContextCapabilities::detect(),
             formats: formats,
             limits: limits,
+            extensions: extensions
         })
     }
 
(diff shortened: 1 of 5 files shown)


kvark (approved):

kvark:
@emilio do you want to also have a look?

Edit: thanks @MortimerGoro ! It's great to finally support packed DS, should get more efficient for our stencil usage.

src/gl_context.rs
```diff
@@ -47,8 +48,10 @@ impl<Native> GLContext<Native>
         };
 
         try!(native_context.make_current());
+        let extensions = gl_.get_string(gl::EXTENSIONS);
+        let extensions: Vec<String> = extensions.split(&[',',' '][..]).map(|s| s.into()).collect();
```
emilio, line 52: Do you need to collect into a `Vec<String>`? I'd expect `Vec<&str>` to work.

src/gl_formats.rs
```diff
@@ -38,18 +42,19 @@ impl GLFormats {
...
         }
     }
 
     #[cfg(target_os="android")]
-    pub fn detect(attrs: &GLContextAttributes, gl: &gl::Gl) -> GLFormats {
+    pub fn detect(attrs: &GLContextAttributes, extensions: &[String]) -> GLFormats {
         // detect if the GPU supports RGB8 and RGBA8 renderbuffer/texture storage formats.
         // GL_ARM_rgba8 extension is similar to OES_rgb8_rgba8, but only exposes RGBA8.
-        let extensions = gl.get_string(gl::EXTENSIONS);
-        let extensions: Vec<&str> = extensions.split(&[',',' '][..]).collect();
```
emilio, line 50: Oh, so yeah, we can :)

src/tests.rs
```diff
@@ -360,3 +360,31 @@ fn test_zero_size() {
...
+    GLContext::<NativeGLContext>::new(size,
+                                      attributes,
+                                      ColorAttachmentType::Texture,
+                                      gl::GlType::default(),
+                                      None).unwrap();
+}
+
+#[test]
+fn test_stencil_no_depth() {
+    let mut attributes = GLContextAttributes::default();
```
emilio: You can use the:

```rust
let attributes = GLContextAttributes {
    depth: true,
    stencil: true,
    .. Default::default(),
};
```

syntax, which is a bit nicer IMO.

emilio (approved):
Looks good % nits, thanks @MortimerGoro!

Also, thanks for taking a look @kvark :)

src/gl_context.rs
MortimerGoro: I didn't use Vec<&str> in this case because the String returned from gl.get_string(gl::EXTENSIONS) is destroyed. We'd need to save it somewhere to have a Vec<&str>

emilio: Oh, didn't notice you were saving all the extensions there, right. Then that looks fine.

PR #98, opened 2017-04-25
Antialias attribute shall not cause a failure to create a WebGLRenderingContext

MortimerGoro:
WebGL spec:

> **The depth, stencil and antialias attributes, when set to true, are requests, not requirements**. The WebGL implementation should make a best effort to honor them. When any of these attributes is set to false, however, the WebGL implementation must not provide the associated functionality. **Combinations of attributes not supported by the WebGL implementation or graphics hardware shall not cause a failure to create a WebGLRenderingContext**. The actual context parameters are set to the attributes of the created drawing buffer. The alpha, premultipliedAlpha and preserveDrawingBuffer attributes must be obeyed by the WebGL implementation.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "offscreen_gl_context"
 license = "MIT / Apache-2.0"
-version = "0.8.7"
+version = "0.8.8"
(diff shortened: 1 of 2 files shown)


jdm:
Thanks!

PR #101, opened 2017-06-01
Bump euclid to 0.14.

nical:
I suppose it would have been fine to require euclid ">= 0.11, <= 0.14" since none of the breaking changes affect this crate (looks like it only uses euclid sizes), but I have a feeling it'll just be simpler if all servo crates require euclid 0.14+.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "offscreen_gl_context"
 license = "MIT / Apache-2.0"
-version = "0.8.10"
+version = "0.9.0"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "offscreen_gl_context"
 license = "MIT / Apache-2.0"
-version = "0.8.10"
+version = "0.9.10"
```
kvark: 0.9.0

kvark (changes requested):

PR #106, opened 2017-09-21
Bump core foundation

glennw:


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "offscreen_gl_context"
 license = "MIT / Apache-2.0"
-version = "0.11.1"
+version = "0.12.0"
```
mbrubeck: It looks like this is a non-breaking change.  If so, this could be 0.11.2.

emilio (approved):

glennw:
@emilio Is bors enabled on this repo?

emilio:
Err, nope

PR #109, opened 2017-10-20
Implement EAGL api for iOS

MortimerGoro:
Implement EAGL api for iOS.  All the tests are passing.

Note: you can use `cargo dinghy test --target x86_64-apple-ios` for easy testing using an iPhone emulator on MacOS

### DIFF
--- a/src/platform/with_eagl/native_gl_context.rs
+++ b/src/platform/with_eagl/native_gl_context.rs
@@ -1,0 +1,140 @@
+use platform::NativeGLContextMethods;
+use GLVersion;
+use gleam::gl;
+use objc::runtime::{BOOL, NO};
+use objc::runtime::{Class, Object};
+use libloading as lib;
+use std::ops::Deref;
+use std::ptr;
+
+type EAGLContext = *mut Object;
+type EAGLSharegroup = *mut Object;
+
+lazy_static! {
+    static ref EAGLCONTEXT_CLASS: &'static Class  = {
+        Class::get("EAGLContext").expect("EAGLContext class not found")
+    };
+
+    static ref OPENGLES_FRAMEWORK: Option<lib::Library>  = {
+        lib::Library::new("/System/Library/Frameworks/OpenGLES.framework/OpenGLES").ok()
+    };
+}
+
+pub struct NativeGLContext(EAGLContext);
+pub type NativeGLContextHandle = NativeGLContext;
+unsafe impl Send for NativeGLContext {}
+
+impl NativeGLContext {
+    fn new_retained(context: EAGLContext) -> Self {
+        let context = unsafe { msg_send![context, retain] };
+        NativeGLContext(context)
+    }
+}
+
+impl Drop for NativeGLContext {
+    fn drop(&mut self) {
+        unsafe {
+            msg_send![self.0, release];
+        }
+    }
+}
+
+impl NativeGLContextMethods for NativeGLContext {
+    type Handle = Self;
+
+    fn get_proc_address(addr: &str) -> *const () {
+        let framework = match *OPENGLES_FRAMEWORK {
+            Some(ref lib) => lib,
+            None => return ptr::null(),
+        };
+        unsafe {
+            let symbol: Result<lib::Symbol<unsafe extern fn()>, _> = framework.get(addr.as_bytes());
+            match symbol {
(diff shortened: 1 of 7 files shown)


src/platform/with_eagl/native_gl_context.rs
```diff
@@ -0,0 +1,128 @@
...
+
+pub struct NativeGLContextHandle(EAGLContext);
+
+unsafe impl Send for NativeGLContextHandle {}
+
+impl NativeGLContextMethods for NativeGLContext {
+    type Handle = NativeGLContextHandle;
+
+    fn get_proc_address(addr: &str) -> *const () {
+        if let Some(ref lib) = *OPENGLES_FRAMEWORK {
```
emilio: nit: maybe:

```rust
let framework = match *OPENGLES_FRAMEWORK {
    Some(ref lib) => lib,
    None => return ptr::null(),
};

// ...
```

So the code is a bit deindented?

src/platform/with_eagl/native_gl_context.rs
```diff
@@ -0,0 +1,128 @@
...
+        0 as *const _
+    }
+
+    fn create_shared(with: Option<&Self::Handle>,
+                     _api_type: &gl::GlType,
+                     api_version: GLVersion) -> Result<NativeGLContext, &'static str> {
+        let context: EAGLContext = unsafe {
+            let context: EAGLContext = msg_send![*EAGLCONTEXT_CLASS, alloc];
+            let version = api_version.major_version() as u32;
+            if let Some(with) = with {
```
emilio: nit: I'd use `match`, but no big deal.

src/platform/with_eagl/native_gl_context.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    fn current() -> Option<Self> {
+        let context: EAGLContext = unsafe {
+            msg_send![*EAGLCONTEXT_CLASS, currentContext]
+        };
+        if context.is_null() {
+            None
+        } else {
+            Some(NativeGLContext(context))
```
emilio: So, this is creating a `NativeGLContext`, which will `release` it when it gets dropped.

However, the call above, and below, don't seem to `addref` it in any way.

So either calling this function will eventually lead to a double-free of the context, or the other two are leaking the context.

src/platform/with_eagl/native_gl_context.rs
```diff
@@ -0,0 +1,128 @@
...
+        if context.is_null() {
+            Err("[EAGLContext initWithAPI] failed")
+        } else {
+            Ok(NativeGLContext(context))
+        }
+    }
+
+    fn is_current(&self) -> bool {
+        let context: EAGLContext = unsafe { 
+            msg_send![*EAGLCONTEXT_CLASS, currentContext]
```
emilio: This could use `current_handle`, to have less `unsafe` code.

emilio (changes requested):
Looks mostly ok, mod nits and the issue below.

src/platform/with_eagl/native_gl_context.rs
MortimerGoro: Yes, that would lead to double-frees or leaked references. I have retained all EAGLContext instances in NativeGLContext and NativeGLContextHandle structs now.

src/platform/with_eagl/native_gl_context.rs
```diff
@@ -0,0 +1,142 @@
...
+                Ok(symbol) => *symbol.deref() as *const (),
+                _ => ptr::null()
+            }
+        }
+    }
+
+    fn create_shared(with: Option<&Self::Handle>,
+                     api_type: &gl::GlType,
+                     api_version: GLVersion) -> Result<NativeGLContext, &'static str> {
+        match *api_type {
```
emilio, line 62: Should this be the other way around? i.e. `if api_type != GlType::Gles` instead of the opposite?

emilio (approved):

src/platform/with_eagl/native_gl_context.rs
emilio: (I know it's equivalent, just maybe a bit simpler)

emilio:
Feel free to merge as soon as travis is green, with or without that :)

emilio:
Travis was green on the previous push, and not much changed....

PR #112, opened 2017-11-14
Update gl_generator

MortimerGoro:
See https://github.com/servo/gleam/pull/130

We also need this to avoid duplicated versions tidy checks to land the new gleam version in Servo.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,14 +1,14 @@
 [package]
 name = "offscreen_gl_context"
 license = "MIT / Apache-2.0"
-version = "0.12.1"
+version = "0.13.0"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -1,14 +1,14 @@
 [package]
 name = "offscreen_gl_context"
 license = "MIT / Apache-2.0"
-version = "0.12.1"
+version = "0.12.2"
```
jdm: We should do a minor version bump, since we use generated bindings in the public API for this crate.

jdm (changes requested):

MortimerGoro:
Done

jdm (approved):

PR #114, opened 2017-12-07
Update euclid to 0.16 and bump version

Eijebong:


kvark:
I'm afraid this is done in #115
Thanks anyway!

PR #123, opened 2018-05-30
Detect more GL limits.

jdm:


src/gl_limits.rs
```diff
@@ -7,19 +7,33 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
...
+    pub max_vertex_uniform_vectors: u32
 }
 
 #[cfg(feature = "serde")]
 impl<'de> Deserialize<'de> for GLLimits {
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
         where D: Deserializer<'de>
     {
-        let values = try!(<[_; 3]>::deserialize(deserializer));
+        let values = try!(<[_; 10]>::deserialize(deserializer));
```
kvark, line 25: any reason why `GLLimits` can't just derive deserialization?

src/gl_limits.rs
```diff
@@ -29,26 +43,87 @@ impl Serialize for GLLimits {
...
+        [self.max_vertex_attribs, self.max_tex_size, self.max_cube_map_tex_size,
+         self.max_combined_texture_image_units, self.max_fragment_uniform_vectors,
+         self.max_renderbuffer_size, self.max_texture_image_units, self.max_varying_vectors,
+         self.max_vertex_texture_image_units, self.max_vertex_uniform_vectors]
             .serialize(serializer)
     }
 }
 
+macro_rules! gl_integer {
+    ($gl:ident, $pname:ident) => {
```
kvark, line 55: what's a particular reason for this to be a macro as opposed to a function?

src/gl_limits.rs
```diff
@@ -29,26 +43,87 @@ impl Serialize for GLLimits {
...
-            gl_.get_integer_v(gl::MAX_CUBE_MAP_TEXTURE_SIZE, &mut limit);
+        let max_vertex_attribs = gl_integer!(gl_, MAX_VERTEX_ATTRIBS);
+        let max_tex_size = gl_integer!(gl_, MAX_TEXTURE_SIZE);
+        let max_cube_map_tex_size = gl_integer!(gl_, MAX_CUBE_MAP_TEXTURE_SIZE);
+        let max_combined_texture_image_units = gl_integer!(gl_, MAX_COMBINED_TEXTURE_IMAGE_UNITS);
+        let max_renderbuffer_size = gl_integer!(gl_, MAX_RENDERBUFFER_SIZE);
+        let max_texture_image_units = gl_integer!(gl_, MAX_TEXTURE_IMAGE_UNITS);
+        let max_vertex_texture_image_units = gl_integer!(gl_, MAX_VERTEX_TEXTURE_IMAGE_UNITS);
+
+        // Based off of https://searchfox.org/mozilla-central/rev/5a744713370ec47969595e369fd5125f123e6d24/dom/canvas/WebGLContextValidate.cpp#523-558
```
kvark, line 91: jeez, I would not want to accept this code if I was reviewing it:
> We are now going to try to read GL_MAX_VERTEX_OUTPUT_COMPONENTS
         * and GL_MAX_FRAGMENT_INPUT_COMPONENTS, however these constants
         * only entered the OpenGL standard at OpenGL 3.2. So we will try
         * reading, and check OpenGL error for INVALID_ENUM.

I think it's a bad idea. If a constant was introduced in GL 3.2, wouldn't it be better for us to just check the context version and go from that?

For an example of what can be wrong with this code: an error could be triggered by something else before our fallible check, and we'd think that it occurred due to the check.

src/gl_limits.rs
jdm: This was an explicit choice a while ago in https://github.com/emilio/rust-offscreen-rendering-context/commit/c97823f97b8fbee8689d5cdaabb22566c03007be (from #57). I'd rather look into that separately from this work.

src/gl_limits.rs
jdm: The debugging output stringifies the name of the enum used.

src/gl_limits.rs
kvark: On a second thought, maybe checking for the context version is not that trivial, since it's returned as string (d'oh, GL).
I still don't like this approach, but I'd not want you to get stuck finding a better way here. At the very least, could we ensure the fallible integer queries expect `GL_INVALID_ENUM` error and nothing else?

src/gl_limits.rs
```diff
@@ -29,26 +43,87 @@ impl Serialize for GLLimits {
...
+                $gl.get_integer_v(gl::$pname, &mut val);
+            }
+            assert_eq!($gl.get_error(), gl::NO_ERROR, "Error retrieving {}", stringify!($pname));
+            val[0] as u32
+        }
+    }
+}
+
+fn gl_fallible_integer(gl_: &gl::Gl, pname: gl::GLenum) -> Result<u32, ()> {
+    let mut val = [0];
```
emilio, line 69: You should probably debug_assert that `gl_.get_error()` is `gl::NO_ERROR` here, otherwise you'll reject known values.

src/gl_limits.rs
```diff
@@ -29,26 +43,87 @@ impl Serialize for GLLimits {
...
 }
 
+macro_rules! gl_integer {
+    ($gl:ident, $pname:ident) => {
+        {
+            let mut val = [0];
+            unsafe {
+                $gl.get_integer_v(gl::$pname, &mut val);
+            }
+            assert_eq!($gl.get_error(), gl::NO_ERROR, "Error retrieving {}", stringify!($pname));
```
emilio, line 62: Maybe just `debug_assert_eq!`, though not big deal :)

emilio (commented):
Heh, I magically started receiving github bugmail again for this repo :)

jdm:
Comments addressed.

src/gl_limits.rs
jdm: I think it's important to catch this in release builds too.

kvark (approved):
Thank you for addressing the concerns!
I'm looking forward to have both GLLimits serialization and fallible integer queries fixed/rewritten one day.

PR #139, opened 2019-06-28
Fix up linking on Windows.

jdm:
This is required for the UWP Servo build.


jdm:
Huh, looks like we never set up Homu on this repo.

PR #140, opened 2019-07-10
Support creating offscreen contexts using ANGLE

jdm:
This ensures that the GL functions are loaded from the appropriate DLL when EGL is in use, and that the right formats are selected based on GL/GLES rather than hardcoded by platform.

### DIFF
--- a/src/gl_context.rs
+++ b/src/gl_context.rs
@@ -35,16 +35,16 @@ impl<Native> GLContext<Native>
                   api_version: GLVersion,
                   shared_with: Option<&Native::Handle>)
                   -> Result<Self, &'static str> {
-        Self::create_shared_with_dispatcher(api_type, api_version, shared_with, None)
+        Self::create_shared_with_dispatcher(&api_type, api_version, shared_with, None)
     }
 
-    pub fn create_shared_with_dispatcher(api_type: gl::GlType,
+    pub fn create_shared_with_dispatcher(api_type: &gl::GlType,
                                          api_version: GLVersion,
                                          shared_with: Option<&Native::Handle>,
                                          dispatcher: Option<Box<GLContextDispatcher>>)
         -> Result<Self, &'static str> {
         let native_context = Native::create_shared_with_dispatcher(shared_with,
-                                                                   &api_type,
+                                                                   api_type,
                                                                    api_version,
                                                                    dispatcher)?;
         let gl_ = match api_type {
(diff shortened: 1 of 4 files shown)


src/gl_context.rs
```diff
@@ -35,16 +35,16 @@ impl<Native> GLContext<Native>
                   api_version: GLVersion,
                   shared_with: Option<&Native::Handle>)
                   -> Result<Self, &'static str> {
-        Self::create_shared_with_dispatcher(api_type, api_version, shared_with, None)
+        Self::create_shared_with_dispatcher(&api_type, api_version, shared_with, None)
```
asajeffrey, line 38: Nit: not sure it's worth making a reference here, copying a GlType shouldn't be that expensive?

src/gl_formats.rs
```diff
@@ -21,62 +21,60 @@ impl GLFormats {
...
-        // On OpenGLES 2 detect via extensions if the GPU supports RGB8 and RGBA8 renderbuffer/texture storage formats.
-        // GL_ARM_rgba8 extension is similar to OES_rgb8_rgba8, but only exposes RGBA8.
-        let has_rgb8 = api_version.major_version() >= 3 || extensions.iter().any(|s| s == "GL_OES_rgb8_rgba8");
-        let has_rgba8 = has_rgb8 || extensions.iter().any(|s| s == "GL_ARM_rgba8");
+            gl::GlType::Gles => {
+                // RGB8 or RGBA8 is guaranteed on OpenGLES 3+.
+                // On OpenGLES 2 detect via extensions if the GPU supports RGB8 and RGBA8 renderbuffer/texture storage formats.
+                // GL_ARM_rgba8 extension is similar to OES_rgb8_rgba8, but only exposes RGBA8.
+                let has_rgb8 = api_version.major_version() >= 3 || extensions.iter().any(|s| s == "GL_OES_rgb8_rgba8");
+                let has_rgba8 = has_rgb8 || extensions.iter().any(|s| s == "GL_ARM_rgba8");
```
asajeffrey, line 52: Gosh this is annoying code, not much we can do about it though!

asajeffrey (approved):
LGTM, not sue why we're passing API types by reference rather than by value, but that's just a matter of taste.

src/gl_context.rs
jdm: Before gleam 0.6.18 the GlType type did not derive Copy.

PR #153, opened 2020-03-25
Fix windows build

asajeffrey:
Gets the windows build to compile.

PR #152, opened 2020-03-25
Fix android build

asajeffrey:
The API for creating EGL conexts changed but the android build didn't keep up.

PR #154, opened 2020-03-25
Fix travis yml

asajeffrey:
Gets the travis build to be green again (includes #152 and #153).


surfman/src/platform/windows/angle/surface.rs
```diff
@@ -158,11 +159,11 @@ impl Device {
             ];
 
             EGL_FUNCTIONS.with(|egl| {
-                let egl_surface = if let Some(HandleOrTexture::Texture(texture)) = share_handle {
+                let egl_surface = if let Some(share_handle) = share_handle {
```
jdm, line 162: These changes don't look like they compile when used with the opexnr webxr backend.

jdm: https://github.com/jdm/surfman/commits/d3d-share is a branch that should address that.

jdm:
If the goal is only to get the build working and the openxr changes will be in a separate PR, that's fine.

asajeffrey:
This is just to get the build working again, atm everything is red on travis, which is suboptimal.

PR #169, opened 2020-04-03
Add a method to resize a widget-backed surface

asajeffrey:
On some platforms (e.g. macos and wayland) widget-backed surfaces don't automatically resize when the widget is resized, so we need an API for that.

### DIFF
--- a/surfman/src/platform/macos/system/surface.rs
+++ b/surfman/src/platform/macos/system/surface.rs
@@ -71,7 +71,9 @@ impl Drop for Surface {
 }
 
 pub(crate) struct ViewInfo {
+    view: NSView,
     layer: CALayer,
+    superlayer: CALayer,
     front_surface: IOSurface,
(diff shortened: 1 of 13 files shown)


surfman/src/platform/macos/system/surface.rs
```diff
@@ -196,6 +201,44 @@ impl Device {
         surface.present()
     }
 
+    /// Resizes a widget surface
+    pub fn resize_surface(&self, surface: &mut Surface, size: Size2D<i32>) -> Result<(), Error> {
+        transaction::begin();
```
jdm: Should we move this transaction code after the next block where we could return with an error?

jdm (approved):

surfman/src/platform/macos/system/surface.rs
asajeffrey: Oh good point.

PR #170, opened 2020-04-04
In angle, don't call glFinish to synchronize surfaces built from D3D11 textures

asajeffrey:
Don't do unnecessary synchronization on surfaces that aren't thread-safe.

### DIFF
--- a/surfman/src/platform/windows/angle/context.rs
+++ b/surfman/src/platform/windows/angle/context.rs
@@ -264,7 +264,7 @@ impl Device {
             Framebuffer::Surface(_) => return Err((Error::SurfaceAlreadyBound, surface)),
         }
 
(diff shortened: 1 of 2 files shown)


surfman/src/platform/windows/angle/context.rs
```diff
@@ -264,9 +264,9 @@ impl Device {
             Framebuffer::Surface(_) => return Err((Error::SurfaceAlreadyBound, surface)),
         }
 
-        // If the surface does not use a DXGI keyed mutex, then finish.
+        // If the surface is synhronized with GLFinish, then finish.
```
lqd:
```suggestion
        // If the surface is synchronized with GLFinish, then finish.
```

jdm (approved):
This is a nice increase in clarity.

PR #176, opened 2020-04-23
Round up `IOSurface` width to the nearest 16-byte boundary if backed by native widgets

pcwalton:
This is required for surfaces to show up on screen.

r? @asajeffrey


asajeffrey:
Sigh. @bors-servo r+

PR #178, opened 2020-05-06
Avoid calling eglMakeCurrent prior to creating a window surface.

### LINKED ISSUE #173: `threads` example displays a black window on X11
The `threads` example in `master` displays a black window on X11. I'm using Arch Linux with a GTX 2080 Ti and the NVIDIA proprietary drivers.

The `offscreen` example works correctly.

Let me know if you need more details.

iamralpht:
If we call eglMakeCurrent with no draw or read surface prior to creating the window surface, then on Nvidia/X11 we get a black window. This change avoids making the context current to read out the EGLConfig used to create the window surface, and instead extracts it directly from the raw EGLContext.

Fixes #173

### DIFF
--- a/surfman/src/platform/unix/x11/surface.rs
+++ b/surfman/src/platform/unix/x11/surface.rs
@@ -4,6 +4,8 @@
 
 use crate::gl::types::{GLenum, GLuint};
 use crate::gl;
+use crate::egl;
+use crate::egl::types::EGLint;
 use crate::platform::generic::egl::context;
 use crate::platform::generic::egl::surface::{EGLBackedSurface, EGLSurfaceTexture};
 use crate::{Error, SurfaceAccess, SurfaceInfo, SurfaceType};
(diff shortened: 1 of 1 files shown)


iamralpht:
@jdm I think this is the smallest change that gets most usages of surfman with a native window working with NVidia EGL. I'm sure it's still possible to fall into the buggy path by creating any other kind of surface ahead of the native window, so that's not very good. However, this is definitely a bug in NVidia's EGL implementation, and I'm not sure how much effort making it impossible to run into is (compared to this change, which is the bare minimum to make the common surfman use cases work).

jdm:
Great detective work! r? @pcwalton who knows this code best.

iamralpht:
FWIW I've also tested this change on Intel graphics (on X11) and it is happy there, too.

hecrj:
I can also confirm this fixes #173 for me! :tada: 

Thank you very much for investigating and fixing this! I can keep exploring new renderers for [`iced`] now!

[`iced`]: https://github.com/hecrj/iced

iamralpht:
I looked at a couple of other libraries that use EGL on various hardware, and another common approach to resolving this issue is to create a 1x1 pbuffer "dummy surface", and make current against that when you want to unlock your main surface.

The [documentation](https://www.khronos.org/registry/EGL/sdk/docs/man/html/eglMakeCurrent.xhtml) says:
> If ctx is not EGL_NO_CONTEXT, then both draw and read must not be EGL_NO_SURFACE unless context is a context which supports being bound without read and draw surfaces. In this case the context is made current without a default framebuffer. The meaning of this is defined by the client API of the supporting context (see chapter 4 of the OpenGL 3.0 Specification, and the GL_OES_surfaceless_context OpenGL ES extension.).

(and then gives some examples of making current with no read/draw buffers, so I'm not sure what to make of that).

Adding a dummy surface would be a bigger change (since it would touch every EGL backend, not just X11), but might result in better compatibility (because with this change, there are still ways a user could call surfman and still get this black window behavior on NVidia).

Please let me know if you'd rather have a change that makes a dummy surface and I'll put it together.

pcwalton:
@bors-servo: r+

Nice find!

PR #180, opened 2020-05-08
Support creating an adapter from a DXGI adapter

jdm:
This is important in cases where an external caller has strict requirements about which adapter should be used when creating a surfman device.


jdm:
The first commit fixes a UWP build error on master.

pcwalton:
@bors-servo: r+

Eventually it would be nice to generalize this to the concept of a "native adapter", but this works.

Issue #181, opened 2020-05-17
Issues with examples on Windows


ArturKovacs:
The `threads` example shows a plain white window and then exits with
```
(exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
```

The  `offscreen` example panics with
```
thread 'main' panicked at 'gl function was not loaded', D:\personal\software\contribution\surfman\target\debug\build\gl-76f8da451bedeb81\out/bindings.rs:20624:13
stack backtrace:
   0: backtrace::backtrace::trace_unsynchronized
             at C:\Users\VssAdministrator\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.40\src\backtrace\mod.rs:66
   1: std::sys_common::backtrace::_print_fmt
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\sys_common\backtrace.rs:77
   2: std::sys_common::backtrace::_print::{{impl}}::fmt
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\sys_common\backtrace.rs:59
   3: core::fmt::write
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libcore\fmt\mod.rs:1052
   4: std::io::Write::write_fmt<std::sys::windows::stdio::Stderr>
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\io\mod.rs:1426
   5: std::sys_common::backtrace::_print
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\sys_common\backtrace.rs:62
   6: std::sys_common::backtrace::print
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\sys_common\backtrace.rs:49
   7: std::panicking::default_hook::{{closure}}
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\panicking.rs:204
   8: std::panicking::default_hook
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\panicking.rs:224
   9: std::panicking::rust_panic_with_hook
             at /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447\/src\libstd\panicking.rs:472
  10: std::panicking::begin_panic<str*>
[...]

PR #182, opened 2020-05-18
Fix for the offscreen example not working on Windows.

ArturKovacs:
Partially addresses #181.

Note that after this fix, the example still doesn't halt but it finishes the main function successfully and produces the png with the colorful triangle.

### DIFF
--- a/surfman/examples/offscreen.rs
+++ b/surfman/examples/offscreen.rs
@@ -80,8 +80,8 @@ fn main() {
     }).unwrap();
     device.bind_surface_to_context(&mut context, surface).unwrap();
 
-    gl::load_with(|symbol_name| device.get_proc_address(&context, symbol_name));
     device.make_context_current(&context).unwrap();
+    gl::load_with(|symbol_name| device.get_proc_address(&context, symbol_name));
 
     let mut pixels: Vec<u8> =
         vec![0; FRAMEBUFFER_WIDTH as usize * FRAMEBUFFER_HEIGHT as usize * 4];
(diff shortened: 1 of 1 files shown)


Brooooooklyn:
also resolved https://github.com/servo/surfman/issues/163

kvark:
@pcwalton is this repository maintained? Is this the right place to file surfman PRs?

surfman/examples/offscreen.rs
```diff
@@ -80,15 +80,17 @@ fn main() {
...
+    gl::load_with(|symbol_name| device.get_proc_address(&context, symbol_name));
 
     let mut pixels: Vec<u8> =
         vec![0; FRAMEBUFFER_WIDTH as usize * FRAMEBUFFER_HEIGHT as usize * 4];
     let tri_vertex_array = TriVertexArray::new(device.gl_api(),
                                                device.surface_gl_texture_target());
 
     unsafe {
+        let surface_info = device.context_surface_info(&context).unwrap().unwrap();
+        gl::BindFramebuffer(gl::FRAMEBUFFER, surface_info.framebuffer_object);
```
jdm, line 93: Are these lines still needed if we move the `device.bind_surface_to_context` on line 81 after the `device.make_context_current` below it?

ArturKovacs: I tried moving the lines {78, 79, 80} after gl::load_with and commenting out gl::BindFramebuffer but that produces an empty output image (ie only transparent pixels).

PR #185, opened 2020-06-02
Context Sharing

### LINKED ISSUE #65: Sharing textures doesn't work on macOS/CGL
Here is what gfx-backend-gl tries to do:
  1. create one GL context ("root") without a surface, most operations are done on that context by rendering into textures.
  2. create an additional GL context for the window/surface. When it's time to present, we want to blit from a renderbuffer of the root context to this context. Then we present the surface.

This is needed in order to detach the "Instance" we have from the "Surface". In order to work properly, it requires that a certain renderbuffer is shared between the GL contexts (rendered into by "root", read by "surface".

We were struggling to get this to work with @zicklag, and we found out that CGL context is [created without any sharing](https://github.com/pcwalton/surfman/blob/ae14cc9afd140bc73e89f3fe8c080f729f09ccd2/surfman/src/platform/macos/cgl/context.rs#L198). I experimented with a surfman API that would allow sharing contexts in [this branch](https://github.com/kvark/surfman/commit/e088d23ce5cf3baba533b4f65ced61d19001bab6#diff-1466eb03dde73fe20bd4ce9df33bd175R206) and confirmed that this would work.

@pcwalton is this a bug in the code, feature not yet implemented (sharing), or a limitation by design?

Note: it is very much desired to avoid the blit altogether, and just be able to render to a surface from the "root" context. But apparently this isn't supported by surfman, at least now. We are looking forward to see if this becomes an option.

kvark:
Closes https://github.com/pcwalton/surfman/issues/65
(API breaking)

### DIFF
--- a/surfman/examples/offscreen.rs
+++ b/surfman/examples/offscreen.rs
@@ -74,7 +74,7 @@ fn main() {
         flags: ContextAttributeFlags::empty(),
     };
     let context_descriptor = device.create_context_descriptor(&context_attributes).unwrap();
-    let mut context = device.create_context(&context_descriptor).unwrap();
+    let mut context = device.create_context(&context_descriptor, None).unwrap();
     let surface = device.create_surface(&context, SurfaceAccess::GPUOnly, SurfaceType::Generic {
         size: Size2D::new(FRAMEBUFFER_WIDTH, FRAMEBUFFER_HEIGHT),
     }).unwrap();
--- a/surfman/examples/threads.rs
+++ b/surfman/examples/threads.rs
@@ -107,7 +107,7 @@ fn main() {
     let context_descriptor = device.create_context_descriptor(&context_attributes).unwrap();
 
     let surface_type = SurfaceType::Widget { native_widget };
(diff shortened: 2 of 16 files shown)


asajeffrey:
This seems like a perfectly reasonable implementation of context sharing, but I'm not sure we should be going back to using context sharing,. Servo used to do that, but it was the source of nasty concurrency bugs. Let's have that discussion over in https://github.com/pcwalton/surfman/issues/65

kvark:
Thanks for taking a look @asajeffrey !
To clarify, this isn't for Servo to go back to context sharing. It will happily keep them non-shared.
This is to be able to use surfman in gfx-backend-gl at all, kind of a blocker.

kvark:
@pcwalton what's the next step here?
If I understand correctly, sharing would be in scope of this library, because all the native platforms it abstracts over support sharing (minus bugs) with the same API. You mentioned on the chat that the library doesn't try to prevent driver bugs, which only seems reasonable to me.
Anything else we need before merging this?

PR #187, opened 2020-06-15
Allow creating a SurfaceTexture from a D3D11 texture

asajeffrey:
For webxr layer management on UWP, life is simpler if we can get surfman to create a `SurfaceTexture` directly from a D3D11 texture. We can't go via a surface annoyingly, since we want the `SurfaceTexture` to be created directly from the EGL pbuffer, not going via `eglCreatePbufferFromClientBuffer`.

### DIFF
--- a/surfman/src/platform/windows/angle/surface.rs
+++ b/surfman/src/platform/windows/angle/surface.rs
@@ -325,7 +325,25 @@ impl Device {
                 } else {
                     None
                 };
+                self.create_surface_texture_from_local_surface(
+                    context,
+                    surface,
+                    local_egl_surface,
+                    local_keyed_mutex,
+                )
+            }
+        })
+    }
 
+    fn create_surface_texture_from_local_surface(
+        &self,
+        context: &Context,
+        surface: Surface,
+        local_egl_surface: EGLSurface,
+        local_keyed_mutex: Option<ComPtr<IDXGIKeyedMutex>>,
+    ) -> Result<SurfaceTexture, (Error, Surface)> {
+        EGL_FUNCTIONS.with(|egl| {
+            unsafe {
                 let _guard = self.temporarily_make_context_current(context);
 
                 GL_FUNCTIONS.with(|gl| {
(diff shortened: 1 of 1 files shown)


asajeffrey:
This is ready for review. cc @jdm @pcwalton

surfman/src/platform/windows/angle/surface.rs
```diff
@@ -368,6 +386,23 @@ impl Device {
         })
     }
 
+    /// Given a D3D11 texture, create a surface texture that wraps that texture. This method is unsafe
+    /// in that the resulting surface is only valid on the current thread.
+    pub unsafe fn create_surface_texture_from_texture(
+        &mut self,
+        context: &mut Context,
+        size: &Size2D<i32>,
+        texture: *mut d3d11::ID3D11Texture2D
```
pcwalton, line 396: Should this be a COM pointer?

pcwalton: If not, document the reference counting semantics in the documentation.

asajeffrey: Looks like I can't make it a COM pointer, since the use case is openxr, and https://docs.rs/openxr/0.12.2/x86_64-pc-windows-msvc/openxr/d3d/enum.D3D11.html exposes `type SwapchainImage = *mut ID3D11Texture2D`. I'll add a comment saying that it is up to the caller to ensure that `texture` is kept live.

jdm (approved):

surfman/src/platform/windows/angle/surface.rs
pcwalton: You could still have made it a COM pointer, because you can wrap raw pointers in COM pointers.

I don't think it's a good idea to subvert the COM `IUnknown` reference counting semantics by requiring the caller to ensure liveness. `surfman` should be maintaining reference counts itself.

asajeffrey: TIL about `ComPtr::from_raw`. I'll submit a PR for that.

PR #190, opened 2020-06-25
Update gl and gl_generator

asajeffrey:
This dedups a dependency in servo.


atouchet:
gl depends on gl_generator so it might be worth updating that to 0.14 in surfman and surfman_android_threads as well.

asajeffrey:
@atouchet good point! Done.

PR #191, opened 2020-06-25
Take D3D11 textures as ComPtrs rather than raw pointers

asajeffrey:
Take D3D11 textures as ComPtrs rather than raw pointers.


asajeffrey:
Addresses https://github.com/servo/surfman/pull/187#discussion_r445658500 r? @pcwalton

Issue #205, opened 2020-11-03
Detach from winit?


kvark:
Winit releases new versions quite often, at least more often than surfman. And it's a dependency of surfman. So once we start depending on surfman in Gecko, there is a problem of dependency duplication that could arise from `winit` (e.g. cocoa stuff). Gecko doesn't like duplicated dependencies because Mercurial is bad at vendoring them.

Would it be possible to somehow avoid depending on winit in `sm-winit` feature? Perhaps, by moving the stable part of winit into a separate crate, or something like that.

kvark:
Actually, it looks like surfman also depends on the macOS crates directly, so it's already in trouble :/
I wonder how Servo team planned to manage the dependency duplication here?

jdm:
We allow duplication on a crate-by-crate basis in https://github.com/servo/servo/blob/071d1d824833dc6103251591e07d8e1bdcea39c0/servo-tidy.toml#L29-L79.

wusyong:
surfman now has a `sm-raw-window-handle` that can feed it with raw window types.
I'm thinking about making an abstraction in servo demo too. So other windowing tools like [glazier](https://github.com/linebender/glazier), or other tool kits like Gtk can use servo too.
But [Connection::from_raw_display_handle](https://github.com/servo/surfman/commit/2fbd4248103cb8640bee28576efaa6784b45121d) is added after 0.6.
Could surfman publish another version if any maintainer has time?

jdm:
Published!

atouchet:
This was done in #271.

PR #206, opened 2020-11-03
Don't depend on winit by default

kvark:
I don't think it's needed here. Adding a feature is always easier for consumers than removing one.
Related to #205


jdm (approved):

kvark:
Replaced by #207

PR #216, opened 2020-12-26
Don't panic if CARGO_CFG_TARGET_FAMILY isn't set

tomaka:
The `CARGO_CFG_TARGET_FAMILY` environment variable isn't defined for non-unix-non-windows platforms.

For context, I'm trying to add support for my custom platform. While I'm obviously not going to try upstream this custom platform, the change in this PR looks positive to me.


jdm:
@bors-servo r+
Thanks!

Issue #208, opened 2020-11-03
Remove euclid dependency


kvark:
It's unclear what value `euclud` brings here. A cursory look reveals only `Point2D` and `Size2D` used, and `surfman` could just roll out its own structures for this.

avr1:
@kvark I'd like to take a look at this, but I'm a little confused as to how to approach the problem. Should I make my own definitions of the Euclid structures, and if so, where in the codebase should I store them and refer all other calls to those structures to? Thanks!

kvark:
@avr1 Sorry! I stopped using surfman in gfx-rs and opted into talking to EGL directly instead, so I can't help here.

avr1:
Thanks anyway! @jdm, do you think I should make my own definitions of the Euclid structures used, and if so, where do you suggest I put them? If not, should we continue to use euclid?

jdm:
Yeah, I would just put them in the root of surfman, or maybe make a geom module that's publicly visible.

avr1:
@jdm Sorry to ask such a newb question, but where exactly is the root of surfman, and how would I access that in the subdirectories? I've been fighting a losing war with the module system.

jdm:
https://github.com/servo/surfman/blob/master/surfman/src/lib.rs

jdm:
From submodules you could refer to `use crate::{Size2D, Point2D};`

PR #218, opened 2020-12-30
Removed euclid dependency, while allowing for more math functions.

avr1:
Resolves issue #208. This is a lot of code, so please review carefully. Thanks!


jdm:
I'm confused what happened here - I thought we were only replacing the uses of Point2D and Size2D in surfman; there should be no need to import all the other unused types from euclid. The code in the examples that uses other euclid types can continue to use euclid; we can change the dependency to a dev-dependency in the Cargo.toml instead.

avr1:
That makes much more sense. Sorry. I'll close this and fix some of the changes.

PR #224, opened 2021-04-22
android: support creating a native widget from a winit window.

iamralpht:
This change allows creating a `surfman::Connection` from a `winit::Window` on the Android platform, using `raw-window-handle` behind the scenes.


iamralpht:
Thanks for the fast turnaround @jdm!

PR #225, opened 2021-04-22
add target dependencies for apple m1

zhengrenzhe:
i build servo on my apple m1 macbook, but got error `error[E0463]: can't find crate for 'objc' ` when compile surfman, so i add the target dependencies for apple m1.

i have already build passed and tested on my m1 macbook with command:

```
cargo build -p surfman
cargo test --lib
```

### DIFF
--- a/surfman/Cargo.toml
+++ b/surfman/Cargo.toml
@@ -2,7 +2,7 @@
 name = "surfman"
 license = "MIT / Apache-2.0"
 edition = "2018"
-version = "0.4.0"
+version = "0.4.1"
 authors = [
     "Patrick Walton <pcwalton@mimiga.net>",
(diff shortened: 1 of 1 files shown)


surfman/Cargo.toml
```diff
@@ -68,6 +68,17 @@ mach = "0.3"
 metal = "0.18"
 objc = "0.2"
 
+[target.aarch64-apple-darwin.dependencies]
```
jdm: Rather than duplicating this list, I recommend using `[target.'cfg(target_os = "macos")'.dependencies]` instead.

jdm (changes requested):

jdm:
@bors-servo r+
Thanks!

PR #233, opened 2021-06-06
Fix #232: android: avoid making EGLContext current without draw/read surfaces when creating window surface.

### LINKED ISSUE #232: surfman with winit crashes on Android Emulator
Using `surfman` with the winit ndk glue results in a crash when creating the window surface on recent Android Emulators, and on cuttlefish (Android's "board" for running the OS in a hypervisor, using virgl).

Surfman's Android platform `create_window_surface` makes the context current without a draw or read buffer (in order to find out the `EGLConfig` to specify when making the window surface), which is only supported on "surfaceless contexts" which the Android virtualized EGL doesn't seem to handle. Making the context current fails, and surfman then crashes a little later on attempting to use a null string returned by `glGetString`.

We can actually just use the `get_context_attr` and `egl_config_from_id` functions to get the `EGLConfig` and avoid making the context current without surfaces.

iamralpht:
This change avoids making the EGLContext current on Android when creating the window surface. The EGLContext implementation provided by the Android Emulator (and by the "cuttlefish" Android board when running in `crosvm`) doesn't support surfaceless contexts which causes making the context current fails in those environments, resulting in a crash later on when we try to use `glGetString`. It's easily possible to get the EGLConfig needed to create the window surface without making the context current, and that's what this change does.

This change has been tested with the winit/ndk integration on the Android Emulator running Android 11 QPR2 on x86_64, and on a Samsung Galaxy Tab S6 (Qualcomm SoC) also running Android 11.

### DIFF
--- a/surfman/Cargo.toml
+++ b/surfman/Cargo.toml
@@ -2,7 +2,7 @@
 name = "surfman"
 license = "MIT / Apache-2.0"
 edition = "2018"
-version = "0.4.2"
+version = "0.4.3"
 authors = [
     "Patrick Walton <pcwalton@mimiga.net>",
     "Emilio Cobos Álvarez <emilio@crisal.io>",
--- a/surfman/src/platform/android/context.rs
+++ b/surfman/src/platform/android/context.rs
@@ -316,11 +316,11 @@ impl Device {
         context::get_proc_address(symbol_name)
     }
 
-    pub(crate) fn context_descriptor_to_egl_config(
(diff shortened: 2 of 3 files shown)


jdm (approved):
Thanks!

iamralpht:
@jdm any chance of an updated release on crates.io? (I'm not sure how those happen for this project...).

jdm:
Published!

iamralpht:
> Published!

Wow, super fast! Thanks!

PR #222, opened 2021-02-26
Update euclid.

emilio:

PR #201, opened 2020-08-01
Update surfman to euclid 0.22

avr1254:

PR #234, opened 2021-08-19
Update euclid to 0.22 and bump version

Bryce-MW:
Updating Euclid is a breaking change which is why I bumped the version. #222 and  #201 are not for the latest version of this repository and should be closed.

I was unable to build the android test so I don't know how well that works but no other changes were needed.

This is part of my effort to update the version of webrender used by Servo. I have made [an issue on the Servo repository](https://github.com/servo/servo/issues/28585) to keep track.

### DIFF
--- a/android-example/rust/Cargo.toml
+++ b/android-example/rust/Cargo.toml
@@ -20,7 +20,7 @@ jni = "0.13"
 log = "0.4"
 
 [dependencies.euclid]
-version = "0.20"
+version = "0.22"
 features = []
 
 [dependencies.surfman]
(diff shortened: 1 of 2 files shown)


Bryce-MW:
It seems that the checks failed only because it couldn't find a runner for Ubuntu 16.04. I don't totally know how GitHub CI works but I could provide a VM if that helps.

yvt:
The Ubuntu 16.04 runner of GitHub Actions has been retired last month (https://github.com/actions/virtual-environments/issues/3287). No idea why it failed earlier than expected. Anyway, #235 should fix the CI issue.

mrobinson:
Landing this change via #244.

PR #236, opened 2022-01-02
bump to winit 0.26

okias:
Compiled on Linux, worked without additional changes.

Signed-off-by: David Heidelberg <david@ixit.cz>


jdm:
Since winit types are exposed in the crate's public API, this needs to be a bump to 0.5. Before we merge this, however, I'd like to see the PRs that update Servo as well, so we don't end up in a situation where bugs in surfman can't be fixed without also getting the winit upgrade merged.

PR #239, opened 2022-08-27
Bump rust nightly to 2022-08-27

becmer:
Fix workflow for 2021 edition on nightly


becmer:
btw. @jdm is there any reason for use of specific version instead of latest one or this is just tech debt?

jdm:
I believe it was pinned to avoid a compiler bug in nightly that was preventing merges.

becmer:
Could you trigger it once again please? It seems that the build is flaky because some networking issues on GH side.

PR #223, opened 2021-03-03
Winit 0.24

pecastro:
Hi.

This is pretty much the same as https://github.com/servo/surfman/pull/179 but just upgrading to v0.24.0

Regards,

    PECastro

PR #246, opened 2023-02-07
Fix linux example to work with winit 0.24

mukilan:
It looks like not all examples were updated when winit was [upgraded](https://github.com/servo/surfman/pull/223/files) from 0.19.1 to 0.24. This PR fixes the code under surfman/examples to build correctly using winit 0.24.

I've tested the 'threads' examples on linux+XWayland but not the 'chaos_game' example as I don't have access to macos. I've also not yet attempted to fix the build errors in the android example.


jdm (approved):

jdm:
Thank you!
@bors-servo r+

PR #247, opened 2023-02-19
Bump winit to 0.28.1

atomgardner:
This fixes the following wayland-client error:

	Attempted to dispatch unknown opcode 0 for wl_shm, aborting.

### DIFF
--- a/android-example/rust/Cargo.toml
+++ b/android-example/rust/Cargo.toml
@@ -5,6 +5,7 @@ authors = ["Patrick Walton <pcwalton@mimiga.net>"]
 edition = "2018"
 
 [lib]
+name = "main"
 crate_type = ["cdylib"]
 
 [profile.dev]
@@ -18,6 +19,7 @@ android_logger = "0.8"
 gl = "0.14"
 jni = "0.13"
 log = "0.4"
+winit = { version = "0.28.1", features = [ "android-native-activity" ] }
 
 [dependencies.euclid]
 version = "0.22"
--- a/surfman/Cargo.toml
+++ b/surfman/Cargo.toml
@@ -44,7 +44,7 @@ version = "0.1"
 optional = true
 
 [dependencies.winit]
-version = "0.24.0"
+version = "0.28.1"
 optional = true
 
 [dependencies.raw-window-handle]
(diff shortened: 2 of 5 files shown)


jdm (approved):

atomgardner:
@jdm, `a45f64c` should fix the Android builds.

mukilan:
@atomgardner : I am able to fix the android build with [these](https://github.com/mukilan/surfman/commit/cd0d9b2247f2d883171ed33ca40028b556cc2acc) changes.

surfman/Cargo.toml
```diff
@@ -89,3 +89,4 @@ winapi = { version = "0.3", features = ["d3d11", "wingdi", "winuser", "libloader
 
 [target.'cfg(target_os = "android")'.dependencies]
 "raw-window-handle" = "0.3.3"
```
mukilan: We should change this to match the version winit uses, so that rust doesn't complain about the HasRawWindowHandle trait not being implemented for Window type.

android-example/rust/Cargo.toml
```diff
@@ -18,6 +19,7 @@ android_logger = "0.8"
 gl = "0.14"
 jni = "0.13"
 log = "0.4"
+winit = { version = "0.28.1", features = [ "android-game-activity" ] }
```
mukilan: Is there any particular reason for using 'android-game-activity' here? The build failure due to missing arm-clang++ seems to go way when using the 'android-native-activity' feature which doesn't require compilation of C++ files.

surfman/Cargo.toml
atomgardner: How did you generate the warning?

android-example/rust/Cargo.toml
atomgardner: @mukilan: The reason is that I misread the `winit` docs. I'll switch this over.

surfman/Cargo.toml
mukilan: It was a rustc error message that I got when I tried building locally without updating the version of raw-window-handle here.

You can build locally using:
```
rustup target add "arm-linux-androideabi"
cd surfman
cargo build --target "arm-linux-androideabi"
```

atomgardner:
@mukilan: nice!

PR #238, opened 2022-08-26
Bump versions

becmer:
winit 0.27.2
euclid 0.22.7
rand 0.8.5

PR #207, opened 2020-11-03
Disable winit by default

kvark:

PR #199, opened 2020-07-29
Update macOS dependencies.

jdm:

PR #249, opened 2023-02-22
Bump version to 0.6.0

mukilan:
Also updated examples to work with winit 0.28.1


atouchet:
If it's not too much of a pain would it be possible to include some dependency updates like in #199 #207 #238 if the version is going to be bumped again?

mukilan:
@atouchet good idea! I'll check which dependencies are easy to upgrade and try a servo build to verify it won't block servo's upgrade of surfman.

jdm (approved):

PR #260, opened 2023-09-19
Fix transparency on macOS

wusyong:
Unlike other platforms, macOS adds another opaque layer on CALayer.
This makes the window opaque even if winit window is set to transparent.

Edit: needs #261 to fix CI.

### DIFF
--- a/surfman/src/platform/macos/cgl/connection.rs
+++ b/surfman/src/platform/macos/cgl/connection.rs
@@ -149,6 +149,7 @@ impl Connection {
         match raw_handle {
             AppKit(handle) => Ok(NativeWidget {
                 view: NSView(unsafe { msg_send![handle.ns_view as id, retain] }),
+                opaque: unsafe { msg_send![handle.ns_window as id, isOpaque] },
             }),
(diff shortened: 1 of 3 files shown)


jdm:
I wonder if we should add a field to NativeWidget that makes these calls configurable? The docs at https://developer.apple.com/documentation/quartzcore/calayer/1410763-opaque suggest that this could have performance implications for the system compositor.

wusyong:
Sure, I can look into adding a method to toggle it. But it'll probably only matter to macOS just like how `glutin` works.
`glutin`'s x11/wayland/Windows transparency is more on `winit`'s config.
I'm on vacation soon. I'll resume the work once I get back next week.

wusyong:
Sorry for the late update. My laptop was broken after vacation and took me a few days to fix.
I added an `opaque` field to the NativeWidget. It's determined by NSWindow's opaque property.
So if the raw window hand or winit window sets the transparency, it will make the layer transparent as well.
Otherwise, normal usage is still the same.

jdm (approved):
Nice!

PR #264, opened 2023-11-11
Update winit to version 0.29

jdm:


atouchet:
winit can also be updated in https://github.com/servo/surfman/blob/master/android-example/rust/Cargo.toml as well.

mrobinson:
I tried to revive this, but I think it might be simpler just to remove winit support in favor of `raw_window_handle` (https://github.com/servo/surfman/pull/271).

@jdm Any thoughts on that approach?

jdm:
Oh, that never even occurred to me. I think that's probably the way to go!

mrobinson:
@jdm thank you!

This is obsoleted by https://github.com/servo/surfman/pull/271.

PR #277, opened 2024-02-26
Load libEGL.so.1 and fallback to unversioned soname

### LINKED ISSUE #276: Consider loading `libEGL.so.1` instead of  `libEGL.so`
After the GStreamer update in [Servo](https://github.com/servo/servo/commit/7fa4ea9740edccc93672d8cc428f6e9d4575a036), the nightly binaries have been [failing](https://github.com/servo/servo/issues/31271) [to launch](https://github.com/servo/servo/issues/31173) on linux distros unless `libegl1-mesa-dev` or equivalent package is installed in the runtime environment.

Based on my [investigation here](https://github.com/servo/servo/pull/31323#issuecomment-1939149784), it seems the root cause of the failure is that surfman is trying to [dynamically open and load](https://github.com/servo/surfman/blob/cc9947d4a805b0903b20c3638e3effd054f54b00/surfman/src/platform/generic/egl/device.rs#L39) functions from `libEGL.so`, but on many systems the symlink from `libEGL.so` to the versioned shared object i.e `libEGL.so.1` doesn't exist unless the `libegl1-mesa-dev` or equivalent package is installed. 

Also, the reason the binaries were loading before the GStreamer update was because the binary had a compile-time link to `libEGL.so.1` (due to gstreamer-sys crate, which on nightly is not present after the GStreamer upgrade since we have also disabled the media stack). Because of this compile time link, the `dlsym` calls are able to succesfully load the functions even though the `dlopen('libEGL.so')` call returns a null handle, indicating failure.

Additionally, @mrobinson found this issue which indicates other bindings like `khronos-egl` [load](https://github.com/timothee-haudebourg/khronos-egl/issues/13#issuecomment-783306616) `libEGL.so.1`  instead of `libEGL.so`.

This issue is to decide if it is a good idea to do the same in surfman. I don't have expertise in EGL, so I am not sure if we still need a fallback to loading `libEGL.so` if `libEGL.so.1` fails to load, like khronos-egl does. I have [tested with a patched surfman](https://github.com/mukilan/servo/actions/runs/7955277302) that loading `libEGL.so.1` does indeed fix the `egl function was not loaded` issues and Servo continues to work.

mukilan:
After the GStreamer update in Servo, the nightly binaries have been failing to launch on linux distros unless `libegl1-mesa-dev` or the equivalent package is installed in the runtime environment.

The binaries were loading before the GStreamer update was because the binary had a compile-time link to libEGL.so.1 due to gstreamer-sys crate. Because of this compile-time link, the dlsym calls are able to succesfully load the functions pointers even though the previous dlopen('libEGL.so') call returned a NULL handle indicating failure.

This patch makes surfman load `libEGL.so.1` first and fallback to `libEGL.so`. If neither are available, then the initialization panics, unlike the previous behavior where we silently succeed if the binary has a link to the libEGL shared object with the symbols for functions that are used at the runtime.

Fixes #276.

### DIFF
--- a/surfman/Cargo.toml
+++ b/surfman/Cargo.toml
@@ -2,7 +2,7 @@
 name = "surfman"
 license = "MIT OR Apache-2.0 OR MPL-2.0"
 edition = "2018"
-version = "0.9.0"
+version = "0.9.1"
 authors = [
     "Patrick Walton <pcwalton@mimiga.net>",
     "Emilio Cobos Álvarez <emilio@crisal.io>",
--- a/surfman/src/platform/generic/egl/device.rs
+++ b/surfman/src/platform/generic/egl/device.rs
@@ -34,12 +34,15 @@ lazy_static! {
 #[cfg(not(any(target_os = "windows", target_os = "macos")))]
 lazy_static! {
     static ref EGL_LIBRARY: EGLLibraryWrapper = {
-        unsafe {
-            EGLLibraryWrapper(dlopen(
-                &b"libEGL.so\0"[0] as *const u8 as *const _,
-                RTLD_LAZY,
-            ))
+        for soname in [b"libEGL.so.1\0".as_ptr(), b"libEGL.so\0".as_ptr()] {
+            unsafe {
+                let handle = dlopen(soname as *const _, RTLD_LAZY);
+                if !handle.is_null() {
+                    return EGLLibraryWrapper(handle);
+                }
+            }
         }
+        panic!("Unable to load the libEGL shared object");
     };
 }
 


surfman/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "surfman"
 license = "MIT OR Apache-2.0 OR MPL-2.0"
 edition = "2018"
-version = "0.9.0"
+version = "0.9.1"
```
mukilan, line 5: I am not sure if we need to bump the minor version here instead, as this patch could break consumers that are working due to the presence of a link to libEGL.so.1 in the binary, similar to servo.

mrobinson (approved):

mukilan:
Thanks for the review!

PR #279, opened 2024-03-15
Reformat code with `cargo fmt`

Nopey:


mrobinson (changes requested):
Great. I noticed that this needed to be done! Would it be possible to make two changes here:

1. Specify a stable rust version in the `rust-toolchain.toml` and include the `rustfmt` component.
2. Add a CI job to run `rustfmt --check` to make sure this doesn't regress.

Thanks!

Nopey:
Added the CI step & pinned rustfmt version.

.github/workflows/main.yml
```diff
@@ -82,3 +82,11 @@ jobs:
       - name: Mark the job as unsuccessful
         if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')
         run: exit 1
+  fmt:
+    name: Rustfmt
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
mrobinson: `checkout@v2` uses a deprecated version of node.

```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/main.yml
```diff
@@ -82,3 +82,11 @@ jobs:
...
+      - uses: mkroening/rust-toolchain-toml@main
+      - name: Rustfmt
```
mrobinson: I think the name is redundant here and you could probably remove it.

mrobinson (changes requested):
Please move the job definition above `build_result` and make sure that `build_result` depends on it. Otherwise, the CI will not fail if `rustfmt --check` fails. 

Otherwise, I have a few nits:

.github/workflows/main.yml
```diff
@@ -82,3 +82,11 @@ jobs:
       - name: Mark the job as unsuccessful
         if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')
         run: exit 1
+  fmt:
+    name: Rustfmt
```
mrobinson: This is a nit but `rustfmt` is the name of the tool.

```suggestion
    name: Run `rustfmt`
```

.github/workflows/main.yml
```diff
@@ -82,3 +82,11 @@ jobs:
       - name: Mark the job as unsuccessful
         if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')
         run: exit 1
+  fmt:
```
mrobinson: No need to abbreviate here:

```suggestion
  format:
```

mrobinson (changes requested):

.github/workflows/main.yml
Nopey: Removed abbreviation, now using `Format:` (with a capital F) to match the Build step's capitalization

mrobinson (approved):

PR #283, opened 2024-04-11
Fix X11 build error

Nopey:
Caught in servo/webxr's CI pipeline https://github.com/servo/webxr/actions/runs/8358606375/job/22880254127

Added the sm-x11 feature to the workflows to ensure it doesn't regress again.


jdm (approved):
Thanks!

PR #284, opened 2024-04-21
Fix some simple clippy lints

jschwe:
Each lint was fixed in an individual commit, for easier review.

After this PR clippy is warning free when run on surfman, examples are excluded and the following lints are allowed:
- clippy::missing_safety_doc
- clippy::unnecessary_cast 
- clippy::result_large_err 
- clippy::mismatched_target_os
- clippy::unnecessary_mut_passed 
- clippy::drop_non_drop

### DIFF
--- a/surfman/build.rs
+++ b/surfman/build.rs
@@ -38,18 +38,18 @@ fn main() {
         || (target_os == "windows" && cfg!(feature = "sm-angle"))
         || target_family.as_ref().map_or(false, |f| f == "unix")
     {
-        let mut file = File::create(&dest.join("egl_bindings.rs")).unwrap();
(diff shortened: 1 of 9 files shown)


mrobinson (approved):
It would also be nice to add a clippy run to GitHub Actions to ensure that these don't regress.

mrobinson (changes requested):
This needs a `cargo fmt` run, I think.

jschwe:
>  It would also be nice to add a clippy run to GitHub Actions to ensure that these don't regress.

I agree - but I think the examples need to be fixed first (they are not compiling, and cause clippy to fail). Locally, I just ignored the issue by deleted the examples folder.

mrobinson (approved):

PR #285, opened 2024-04-22
Add initial support for OpenHarmony

jschwe:
This adds support all the surfman APIs servo seems to be requiring on OpenHarmony.
Generic surfaces are not supported at the moment.
This patch is heavily based on the existing android implementation.

### DIFF
--- a/surfman/build.rs
+++ b/surfman/build.rs
@@ -15,8 +15,9 @@ fn main() {
         windows: { target_os = "windows" },
         macos: { target_os = "macos" },
         android: { target_os = "android" },
+        ohos: { target_env = "ohos" },
         // TODO: is `target_os = "linux"` the same as the following check?
-        linux: { all(unix, not(any(macos, android))) },
+        linux: { all(unix, not(any(macos, android, ohos))) },
 
         // Features:
         // Here we collect the features that are only valid on certain platforms and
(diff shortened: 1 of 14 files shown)


surfman/build.rs
```diff
@@ -15,8 +15,9 @@ fn main() {
         windows: { target_os = "windows" },
         macos: { target_os = "macos" },
         android: { target_os = "android" },
+        ohos: { target_env = "ohos" },
         // TODO: is `target_os = "linux"` the same as the following check?
-        linux: { all(unix, not(any(macos, android))) },
+        linux: { all(unix, not(any(macos, android, ohos))) },
```
jschwe, line 20: I believe it might be a good opportunity to rename the `linux` cfg option to something that makes it more clear, that it is meant for unix systems with x11 or wayland. But I'm not sure what a good name would be.

wusyong:
Could it just reuse the same Android implementation and add `#[cfg(ohos)]` when it's necessary?

surfman/build.rs
wusyong: I think this script is just reusing what winit has. Not many other UNIX-like OSes care about desktop environments usually. It's okay to stay linux IMHO.

jschwe: If it is indeed based on winit, then winit has since changed the names - there is now e.g. `free_unix`, `x11` and `wayland`. 
https://github.com/rust-windowing/winit/blob/master/build.rs
I think it would make sense to rename our cfg flags here to match the ones from `winit` - would you agree?

wusyong: Oh I didn't realize they changed it. It seems like a good idea!

jschwe: I opened a seperate PR for the `cfg` changes in https://github.com/servo/surfman/pull/290

jschwe:
> Could it just reuse the same Android implementation and add #[cfg(ohos)] when it's necessary?

Hmm, perhaps the common parts could be extracted to a shared `egl` implementation.

jschwe:
@mrobinson Do you have any opinions regarding this PR and https://github.com/servo/surfman/pull/290? Would you prefer if the android and ohos implementations would be merged into one directory?

mrobinson:
> @mrobinson Do you have any opinions regarding this PR and #290? Would you prefer if the android and ohos implementations would be merged into one directory?

Sorry for taking so long to look at this. I think that if both platforms use EGL and they have a lot of shared code it probably makes sense to make a shard EGL backend, as you say.

jschwe:
Superceded by #293

PR #293, opened 2024-05-24
Add initial support for OpenHarmony

jschwe:
Follow-up PR to #285.

Todo: 
- Is platform/egl really the best name for the shared code?
- It would be great if someone else could test the changes on android - I couldn't get the servo android build to work on my personal machine (unrelated to these changes)

### DIFF
--- a/surfman/src/platform/egl/ohos_ffi.rs
+++ b/surfman/src/platform/egl/ohos_ffi.rs
@@ -28,6 +28,15 @@ extern "C" {
     /// Can be used to query information like height and width.
     /// See the official [Documentation] for detailed usage information.
     ///
+    /// # Safety
+    ///
+    /// The `window` handle must be valid.
+    /// The variable arguments which must be passed to this function vary depending on the
+    /// value of `code`.
+    /// For `NativeWindowOperation::GET_BUFFER_GEOMETRY` the function two output i32 pointers
+    /// `height: *mut i32` and `width: *mut i32` must be passed as variadic arguments.
+    ///
+    ///
     /// [Documentation]: <https://gitee.com/openharmony/docs/blob/master/en/application-dev/reference/apis-arkgraphics2d/_native_window.md>
     pub(crate) fn OH_NativeWindow_NativeWindowHandleOpt(
         window: *mut OHNativeWindow,
--- a/surfman/src/platform/egl/ohos_ffi.rs
+++ b/surfman/src/platform/egl/ohos_ffi.rs
@@ -1,0 +1,37 @@
+#![allow(non_camel_case_types)]
+#![allow(non_upper_case_globals)]
+#![allow(non_snake_case)]
+#![allow(unused)]
+
+#[repr(C)]
+pub struct NativeWindow {
+    _unused: [u8; 0],
+}
+
+pub type OHNativeWindow = NativeWindow;
+
+#[repr(transparent)]
+pub(crate) struct NativeWindowOperation(core::ffi::c_int);
+
+impl NativeWindowOperation {
+    pub const GET_BUFFER_GEOMETRY: Self = Self(1);
+}
+
+/// According to the [native window guidelines], users need to link against
+/// both the NDK and `native_window`.
+/// [native window guidelines]: <https://gitee.com/openharmony/docs/blob/master/en/application-dev/graphics/native-window-guidelines.md>
+#[link(name = "native_window")]
+#[link(name = "ace_ndk.z")]
+extern "C" {
+    /// Sets or obtains the attributes of a native window
+    ///
+    /// Can be used to query information like height and width.
+    /// See the official [Documentation] for detailed usage information.
+    ///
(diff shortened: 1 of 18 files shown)


mrobinson:
This is looking great. I tested the Android build locally and it compiled. I didn't test it in an emulator.

I suppose this is just blocked on a new raw-window-handle release now.

jschwe:
> I suppose this is just blocked on a new raw-window-handle release now.

Ah, actually raw-window-handle 0.6.2. is already released, I just forgot to update this PR after the release.
I adjusted the PR to use rwh 0.6.2 and added the missing code for ohos.

surfman/src/platform/egl/surface/mod.rs
```diff
@@ -0,0 +1,87 @@
+// surfman/surfman/src/platform/egl/android_surface.rs
```
mrobinson: I think this file name is incorrect now. To be honest, I'm not sure why the file name is included at the top of these files. It seems redundant.

surfman/src/platform/egl/surface/ohos_surface.rs
```diff
@@ -0,0 +1,280 @@
...
+        &mut self,
+        context: &Context,
+        native_widget: NativeWidget,
+    ) -> Result<Surface, Error> {
+        let mut height: i32 = 0;
+        let mut width: i32 = 0;
+        // Safety: `OH_NativeWindow_NativeWindowHandleOpt` takes two output i32 pointers as
+        // variable arguments when called with `GET_BUFFER_GEOMETRY`. See the OHNativeWindow
+        // documentation for details:
+        // https://gitee.com/openharmony/docs/blob/master/en/application-dev/reference/apis-arkgraphics2d/_native_window.md
```
mrobinson: Perhaps this should move to a rustdoc safety section?

mrobinson (changes requested):
Just a couple small comments:

surfman/src/platform/egl/surface/ohos_surface.rs
jschwe: ~~This comment is intended for reviewers of the unsafe block, since `OH_NativeWindow_NativeWindowHandleOpt` takes different arguments depending on the value of second parameter, and thus needs to be manually cross-referenced with the documentation during review.
Callers of this function don't need to know about `OH_NativeWindow_NativeWindowHandleOpt`.~~

To be honest I'm not really sure what the Safety preconditions of `create_window_surface()` should be (and if it needs to be unsafe). But I'm pretty sure the Safety preconditions should be the same as the android version (which isn't documented either, though)

surfman/src/platform/egl/surface/mod.rs
jschwe: Yeah, I was also wondering about this, but assumed that perhaps there was some kind of tooling I don't know about, which requires these comments. 

I will fix the name here, but if there is no reason for these comments (anymore), then I'd also be happy to open a PR that removes all of the filenames at the top.

surfman/src/platform/egl/surface/ohos_surface.rs
jschwe: I added a Safety section to the rustdoc of the function declaration in `ohos_ffi.rs`

mrobinson: I agree with your original statement and IMO the original placement isn't an issue. I've just seen that clippy often complains if an unsafe function does not have a "# Safety" section. It just seems that since you have already gone through the trouble of writing a nice comment here, we can use it to make clippy happy as well. :D

surfman/src/platform/egl/ohos_ffi.rs
```diff
@@ -0,0 +1,46 @@
...
+    /// Can be used to query information like height and width.
+    /// See the official [Documentation] for detailed usage information.
+    ///
+    /// # Safety
+    ///
+    /// The `window` handle must be valid.
+    /// The variable arguments which must be passed to this function vary depending on the
+    /// value of `code`.
+    /// For `NativeWindowOperation::GET_BUFFER_GEOMETRY` the function two output i32 pointers
+    /// `height: *mut i32` and `width: *mut i32` must be passed as variadic arguments.
```
mrobinson:
```suggestion
    ///  - The `window` handle must be valid.
    ///  - The variable arguments which must be passed to this function vary depending on the
    ///    value of `code`.
    ///  - For `NativeWindowOperation::GET_BUFFER_GEOMETRY` the function takes two output
    ///    i32 pointers, `height: *mut i32` and `width: *mut i32` which are passed as variadic
    ///    arguments.
```

mrobinson (approved):

mrobinson:
Is this still a draft or is it okay to merge this?

jschwe:
It's okay to merge from my side. Should I squash the review changes into the second commit, or will you squash everything into one commit anyway?

mrobinson:
The merge queue automatically squashes commits, so it shouldn't be an issue. Thanks!

PR #295, opened 2024-06-10
Bump version to 0.9.4

jschwe:


jschwe:
@mrobinson Could you perhaps create a new crates.io release? This is basically the last dependency bump blocking me

mrobinson:
This look great! Sorry for not releasing earlier. To land we'll just need to fix the Windows CI like we recently did for Servo (essentially just upgrading LLVM with choco). I can do that and also land this change tomorrow.

jschwe:
Hmm, I added your fix from the main servo repo here, but it seems that the build is stuck on windows. Maybe we can retry tomorrow and see if it gets stuck consistently.

Edit: Still gets stuck on the build-script-build of mozangle

mrobinson:
I've gotten a slightly different version of the workaround functioning here: https://github.com/servo/surfman/pull/296

jschwe:
Great! How about you also bump the version on your PR, and I'll just close mine?

mrobinson:
> Great! How about you also bump the version on your PR, and I'll just close mine?

I'm happy to keep them as two separate PRs since they do different things. Then if we ever need to revert one, we can keep the other.

jschwe:
Okay, dropped the workaround commit on my end

mrobinson (approved):

mrobinson:
I've published the release: https://crates.io/crates/surfman/0.9.4

jschwe:
Thanks!

PR #299, opened 2024-07-15
chore: Fix Windows workflow

dklassic:
Windows workflow currently breaks due to chocolatey return failure when updating llvm to a specific version 17.0.6 but llvm 18.1.6 is already installed. (see https://github.com/servo/surfman/actions/runs/9890578015/job/27369196314#step:4:19)

We should maybe just skip upgrading (downgrading?) it for now.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -66,17 +66,17 @@ jobs:
       if: matrix.target != 'default' && startsWith(matrix.target, 'aarch64-uwp-windows-msvc') != true
       run: |
         rustup +${{steps.toolchain.outputs.name}} target add ${{ matrix.target }}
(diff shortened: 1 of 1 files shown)


jdm (approved):

mrobinson:
For some reason the Windows build isn't terminating with these changes.

mrobinson (commented):
Try upgrading LLVM

mrobinson (commented):
Try upgrading LLVM

mrobinson:
I ended up having to downgrade LLVM here instead of upgrading, otherwise this would stall on the mozangle build. In general, the mozangle build is not great. It takes a very long time and this issue with LLVM is a bad sign. I hope to find some time to debug it soon.

PR #300, opened 2024-07-25
Fix error when creating WARP device on Windows

crbrz:
`D3D11CreateDevice` doesn't need an adapter when driver type is `D3D_DRIVER_TYPE_WARP`. 

Fixes the error reported in https://github.com/servo/servo/issues/26431 and enables running multiple WPT tests in parallel on Windows.


jdm (approved):
Exciting! Thanks for fixing this.

wusyong (approved):

PR #303, opened 2024-08-03
Unbreak publishing

jdm:
Fixes I made in order to publish 0.9.5.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,7 @@ sparkle = { version = "0.1", optional = true }
 osmesa-sys = { version = "0.1", optional = true }
 rwh_05 = { package = "raw-window-handle", version = "0.5.2", features = ["std"], optional = true }
 rwh_06 = { package = "raw-window-handle", version = "0.6.2", features = ["std"], optional = true }
+serial_test = "3.1.0"
 
 [dev-dependencies]
 clap = "2"
@@ -49,7 +50,6 @@ gl = "0.14"
 png = "0.17"
 rand = "0.8"
 winit = "0.29"
-serial_test = "3.1.0"
 
 [target.'cfg(target_os = "macos")'.dependencies]
 cgl = "0.3.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,6 @@ sparkle = { version = "0.1", optional = true }
 osmesa-sys = { version = "0.1", optional = true }
 rwh_05 = { package = "raw-window-handle", version = "0.5.2", features = ["std"], optional = true }
 rwh_06 = { package = "raw-window-handle", version = "0.6.2", features = ["std"], optional = true }
-serial_test = "3.1.0"
 
 [dev-dependencies]
 clap = "2"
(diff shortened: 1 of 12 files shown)


mrobinson:
Sorry for breaking things and thanks for the fix. Let me know when this is ready and I'll give it a review.

Cargo.toml
```diff
@@ -42,14 +42,14 @@ sparkle = { version = "0.1", optional = true }
 osmesa-sys = { version = "0.1", optional = true }
 rwh_05 = { package = "raw-window-handle", version = "0.5.2", features = ["std"], optional = true }
 rwh_06 = { package = "raw-window-handle", version = "0.6.2", features = ["std"], optional = true }
+serial_test = "3.1.0"
```
jschwe: Why does serial_test need to be added to the main packages dependencies?

mukilan: It is to allow the compilation of android_example. It imports the test module from the main library in `android-example/rust/src/lib.rs:17` for implementation of JNI methods.

mukilan:
I've fixed some more errors in android-example (crate-type & edition in android-example/rust/Cargo.toml). The rust compilation fails at the final link step. That might need more investigation. @jdm was there anything else you wanted to address or is this in a state to be merged?

Cargo.toml
jschwe: If `serial_test` is required by the example, then shouldn't it be added to the Cargo.toml of the example crate? 
Adding a mod via `path = ` shouldn't be affected by the Cargo.toml of the foreign crate. Not sure what I'm missing.

mukilan: android-example doesn't use serial_test directly. It [uses](https://github.com/servo/surfman/blob/b6a2125cee6fda6259e8240d4338b0c55a357a0d/android-example/rust/src/lib.rs#L17) the `test` module of the surfman crate. It is the test module in surfman that uses the serial_test, so I don't think adding it to android-example's manifest will work.

However, I think another approach to eliminate this dependency would be to conditionally import serial_test in surfman/test.rs using `cfg(not(feature = "sm-test"))` and similarly use `cfg_attr(not(feature="sm-test"), serial)` on all the test methods.

jschwe: I see. I believe we should at the very least make the `serial_test` dependency optional and move it behind a feature flag, since there is no need to pull in `serial_test` when compiling servo.

mukilan: In [5ae1ac4](https://github.com/servo/surfman/pull/303/commits/5ae1ac49f6e0f282f81a467f181e894214785b78), I've made the use of `#[serial]` conditional on `not(feature = "sm-test")` and I've move serial_test back to dev-dependencies.

android-example/app/src/main/jniLibs/arm64-v8a/libsurfman_android_threads.so
jdm, line 1: I was hoping to add a working symlink for this library but couldn't complete a build.

mukilan: I see. I can look into this separately. I'll merge this PR for now to allow publishing of the crate.

mukilan (approved):

PR #305, opened 2024-08-15
Make it possible to force GLES usage on Linux

webbeef:


jdm:
Have you verified that this actually works? For example, there's another instance of this in device.rs in the same directory.

webbeef:
Yes, on a Pinephone Pro this worked with the initial patch, but I updated it with your suggestion.

jdm (approved):
This seems like a reasonable quick fix to expose the functionality.

webbeef:
Thanks! Do you plan to do a new release in the near future?

jdm:
Yeah, I'll publish 0.9.6 once this merges.

PR #307, opened 2024-08-19
Ensure bytes per row value meets alignment requirements for macOS surfaces

jdm:
Fixes the crash observed in https://github.com/servo/servo/issues/29809. The new FFI declaration belongs in https://github.com/servo/core-foundation-rs/blob/main/io-surface/src/lib.rs#L162, but I'm putting this up now in case anybody else hit this.

### DIFF
--- a/src/platform/macos/system/surface.rs
+++ b/src/platform/macos/system/surface.rs
@@ -15,7 +15,7 @@ use cocoa::quartzcore::{transaction, CALayer, CATransform3D};
 use core_foundation::base::TCFType;
 use core_foundation::dictionary::CFDictionary;
 use core_foundation::number::CFNumber;
-use core_foundation::string::CFString;
+use core_foundation::string::{CFString, CFStringRef};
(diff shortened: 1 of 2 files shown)


src/platform/macos/system/surface.rs
```diff
@@ -319,6 +324,7 @@ impl Device {
         };
 
         unsafe {
+            let bytesPerRow = IOSurfaceAlignProperty(kIOSurfaceBytesPerRow, (size.width * BYTES_PER_PIXEL) as usize) as i32;
```
mrobinson:
```suggestion
            let bytes_per_row = IOSurfaceAlignProperty(kIOSurfaceBytesPerRow, (size.width * BYTES_PER_PIXEL) as usize) as i32;
```

mrobinson (approved):

mrobinson:
Nice!

mrobinson (approved):

PR #308, opened 2024-08-23
Update toolchain and cfg_aliases to match servo

jschwe:
* Update cfg_aliases to fix the new unexpected_cfgs lint
* Update the rust_toolchain to match servo and prevent warning about -Zcheck-cfg


mrobinson (approved):
Looks fine to me. I don't see any current dependents on surfman on `crates.io` so this seems like a safe change.

PR #311, opened 2024-08-28
Use cstr literals where possible

jschwe:
Convert usage of byte str and normal str with a `\0` to cstr literals where possible

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "surfman"
 license = "MIT OR Apache-2.0 OR MPL-2.0"
-edition = "2018"
+edition = "2021"
 version = "0.9.8"
 authors = [
     "Patrick Walton <pcwalton@mimiga.net>",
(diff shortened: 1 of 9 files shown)


mrobinson:
This seems fine though it seems the build is failing on Windows.

jdm:
I suspect c strings are i8 on windows? It's a common cross-platform error.

jschwe:
> I suspect c strings are i8 on windows? It's a common cross-platform error.

The nice thing about `CStr` is that the `as_ptr()` method documents that it [returns `*const c_char`]（https://doc.rust-lang.org/stable/std/ffi/struct.CStr.html#method.as_ptr. This allows us to remove a lot of the casts in the window specific code.

While removing those needless casts, I stumbled a bit over [CreateWindowExA](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-createwindowexa) accepting a `u16` encoded in a pointer type and added a comment in the source code explaining it.

jdm (approved):
These changes are great! Thanks for cleaning up the code.

PR #313, opened 2024-08-31
replace implementation module with declarative macro

mukilan:
The current approach of having a free standing module that is instantiated multiple times with the explicit `path` often causes confusion when reading the code. So replace this with a very simple declarative macro that hopefully is more explicit and simpler than the current approach.


wusyong (approved):
Really love this move. I also stuck too many times trying to trace the code.

jdm (approved):
Seems reasonable!

PR #321, opened 2024-10-17
Use `glow` instead of `gl_generator` for GL/GLES bindings

sagudev:
See https://github.com/servo/servo/issues/33539.

Note that this PR will expose glow types in surfmans public API.

### DIFF
--- a/android-example/app/src/androidTest/java/org/mozilla/surfmanthreadsexample/SurfmanInstrumentedTest.java
+++ b/android-example/app/src/androidTest/java/org/mozilla/surfmanthreadsexample/SurfmanInstrumentedTest.java
@@ -23,7 +23,7 @@ public class SurfmanInstrumentedTest {
     private static native void testDeviceCreation();
     private static native void testGenericSurfaceCreation();
     private static native void testGL();
-    private static native void testNewlyCreatedContextsAreNotCurrent();
+    private static native void testNewlyCreatedContextsAreCurrent();
     private static native void testSurfaceTextureBlitFramebuffer();
     private static native void testSurfaceTextureRightSideUp();
 
@@ -75,8 +75,8 @@ public class SurfmanInstrumentedTest {
     }
 
     @Test
-    public void newlyCreatedContextsAreNotCurrent() {
-        testNewlyCreatedContextsAreNotCurrent();
+    public void newlyCreatedContextsAreCurrent() {
+        testNewlyCreatedContextsAreCurrent();
     }
 
     @Test
--- a/examples/offscreen.rs
+++ b/examples/offscreen.rs
@@ -129,7 +129,10 @@ fn main() {
 
     unsafe {
         let surface_info = device.context_surface_info(&context).unwrap().unwrap();
-        gl::BindFramebuffer(gl::FRAMEBUFFER, surface_info.framebuffer_object);
+        gl::BindFramebuffer(
+            gl::FRAMEBUFFER,
+            surface_info.framebuffer_object.map_or(0, |fbo| fbo.0.get()),
+        );
         gl::Viewport(0, 0, FRAMEBUFFER_WIDTH, FRAMEBUFFER_HEIGHT);
         ck();
         gl::ClearColor(0.0, 0.0, 0.0, 1.0);
@@ -200,7 +203,7 @@ impl TriVertexArray {
                 gl::FLOAT,
                 gl::FALSE,
                 12,
-                0 as *const GLvoid,
+                0 as _,
             );
             ck();
             gl::VertexAttribPointer(
@@ -209,7 +212,7 @@ impl TriVertexArray {
                 gl::UNSIGNED_BYTE,
                 gl::TRUE,
                 12,
-                8 as *const GLvoid,
+                8 as _,
             );
             ck();
             gl::EnableVertexAttribArray(tri_program.position_attribute as GLuint);
--- a/examples/threads.rs
+++ b/examples/threads.rs
@@ -285,12 +285,17 @@ impl App {
         unsafe {
             self.device.make_context_current(&self.context).unwrap();
 
-            let framebuffer_object = match self.device.context_surface_info(&self.context) {
-                Ok(Some(surface_info)) => surface_info.framebuffer_object,
-                _ => 0,
-            };
+            let framebuffer_object = self
+                .device
+                .context_surface_info(&self.context)
+                .ok()
+                .flatten()
+                .and_then(|surface_info| surface_info.framebuffer_object);
 
-            gl::BindFramebuffer(gl::FRAMEBUFFER, framebuffer_object);
+            gl::BindFramebuffer(
+                gl::FRAMEBUFFER,
+                framebuffer_object.map_or(0, |fbo| fbo.0.get()),
+            );
             gl::Viewport(0, 0, self.window_size.width, self.window_size.height);
 
             gl::ClearColor(0.0, 0.0, 1.0, 1.0);
@@ -400,7 +405,8 @@ impl App {
             gl::BindTexture(
                 self.device.surface_gl_texture_target(),
                 self.device
-                    .surface_texture_object(self.texture.as_ref().unwrap()),
+                    .surface_texture_object(self.texture.as_ref().unwrap())
+                    .map_or(0, |tex| tex.0.get()),
             );
             gl::Uniform1i(self.blit_vertex_array.blit_program.source_uniform, 0);
             ck();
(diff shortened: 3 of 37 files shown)


sagudev:
So the problem is that:
```rust
thread_local! {
    pub static GL_FUNCTIONS: Gl = unsafe {Gl::from_loader_function(context::get_proc_address)};
}
```
does not work because glow requires for context to be current, because it does some querying (version, extension). My idea is to simply replace all usages of `GL_FUNCTIONS` with `let gl = unsafe {Gl::from_loader_function(context::get_proc_address)};` as most of the time we are already current in the calls. This would make querying every time, but functions are not hot AFAIK, so this should be ok. Alternative would be to have Gl stored on context, but this needs to be lazy value, as context creation does not make context current.

sagudev:
I am wondering do we really need newly-created contexts that not immediately made current, or can we just made them current on creation? Linked issue on test is about something different. This was introduced in https://github.com/servo/surfman/commit/dae72e4106028b002d5d41ed4bbe750d6c1b7d96. cc @pcwalton @jdm

jdm:
Original issue about inconsistent behavior: https://github.com/pcwalton/surfman/issues/7

jdm:
I believe we could change the behavior to always make new contexts current.

sagudev:
This is very close now, macos fails due to not clearing gl error before running create surface. Needs some cleanup (warnings) and updated servo companion PR.

jdm:
If you rebase this branch on top of https://github.com/servo/surfman/pull/326/ then the crash in https://github.com/servo/servo/pull/34328 will go away \o/

mrobinson:
Is this ready to be reviewed now?

sagudev:
> Is this ready to be reviewed now?

Yes

.gitignore
```diff
@@ -14,3 +14,4 @@ gc.properties
 /android-example/**/.settings
 SDL2
 *.code-workspace
+.vscode/settings.json
```
mrobinson: This should probably be a separate change and likely should be everything in the `.vscode` directory.

android-example/app/src/androidTest/java/org/mozilla/surfmanthreadsexample/SurfmanInstrumentedTest.java
```diff
@@ -76,7 +76,7 @@ public void gl() {
 
     @Test
     public void newlyCreatedContextsAreNotCurrent() {
-        testNewlyCreatedContextsAreNotCurrent();
+        testNewlyCreatedContextsAreCurrent();
     }
```
mrobinson: This seems to have the opposite meaning now, but the function name `newlyCreatedContextsAreNotCurrent` is the same.

examples/threads.rs
```diff
@@ -287,10 +287,13 @@ impl App {
 
             let framebuffer_object = match self.device.context_surface_info(&self.context) {
                 Ok(Some(surface_info)) => surface_info.framebuffer_object,
-                _ => 0,
+                _ => None,
             };
```
mrobinson: This could use `map` and `ok` now.

examples/threads.rs
```diff
@@ -287,10 +287,13 @@ impl App {
...
 
-            gl::BindFramebuffer(gl::FRAMEBUFFER, framebuffer_object);
+            gl::BindFramebuffer(
+                gl::FRAMEBUFFER,
+                framebuffer_object.map(|fbo| fbo.0.get()).unwrap_or(0),
```
mrobinson: Perhaps use `map_or` here and other places this kind of thing is done.

examples/offscreen.rs
```diff
@@ -129,7 +129,13 @@ fn main() {
...
     unsafe {
         let surface_info = device.context_surface_info(&context).unwrap().unwrap();
-        gl::BindFramebuffer(gl::FRAMEBUFFER, surface_info.framebuffer_object);
+        gl::BindFramebuffer(
+            gl::FRAMEBUFFER,
+            surface_info
+                .framebuffer_object
+                .map(|fbo| fbo.0.get())
+                .unwrap_or(0),
+        );
```
mrobinson: This could use `map_or`.

src/context.rs
```diff
@@ -79,29 +78,18 @@ pub(crate) fn current_context_uses_compatibility_profile(_gl: &Gl) -> bool {
...
-                let extension = gl.GetStringi(gl::EXTENSIONS, extension_index) as *const c_char;
-                let extension = CStr::from_ptr(extension);
-                if extension.to_str() == Ok("GL_ARB_compatibility") {
-                    return true;
-                }
-            }
-        }
-
-        false
+        gl.supported_extensions().contains("GL_ARB_compatibility")
```
mrobinson, line 93: Nice!

src/info.rs
```diff
@@ -38,28 +35,10 @@ impl GLVersion {
...
-                .parse()
-                .expect("Couldn't parse the minor GL version!");
-            GLVersion {
-                major: major_version,
-                minor: minor_version,
-            }
+        let version = gl.version();
+        Self {
+            major: version.major as u8,
+            minor: version.minor as u8,
```
mrobinson, line 41: Excellent to see this.

src/platform/egl/surface/android_surface.rs
```diff
@@ -379,7 +377,7 @@ impl Device {
     ///
     /// This will be `GL_TEXTURE_2D` or `GL_TEXTURE_RECTANGLE`, depending on platform.
     #[inline]
-    pub fn surface_gl_texture_target(&self) -> GLenum {
+    pub fn surface_gl_texture_target(&self) -> u32 {
```
mrobinson, line 380: Gleam still has an `GLenum` type and it's defined as `c_uint`, so I'm curious why this is changing here? https://docs.rs/gleam/latest/gleam/gl/types/type.GLenum.html

src/platform/generic/egl/surface.rs
```diff
@@ -93,39 +93,36 @@ impl EGLBackedSurface {
 
         unsafe {
             // Create our texture.
-            let mut texture_object = 0;
-            gl.GenTextures(1, &mut texture_object);
+            let texture_object = gl.create_texture().unwrap();
```
mrobinson: It would be nice to have error handling one day.

src/platform/macos/cgl/context.rs
```diff
@@ -125,6 +123,16 @@ impl Clone for ContextDescriptor {
     }
 }
 
+fn make_current(cgl_context: CGLContextObj) -> Result<(), Error> {
+    unsafe {
+        let err = CGLSetCurrentContext(cgl_context);
+        if err != kCGLNoError {
+            return Err(Error::MakeCurrentFailed(err.to_windowing_api_error()));
```
mrobinson, line 130:
```suggestion
        if CGLSetCurrentContext(cgl_context) != kCGLNoError {
            return Err(Error::MakeCurrentFailed(err.to_windowing_api_error()));
```

src/platform/macos/cgl/context.rs
```diff
@@ -125,6 +123,16 @@ impl Clone for ContextDescriptor {
     }
 }
 
+fn make_current(cgl_context: CGLContextObj) -> Result<(), Error> {
```
mrobinson:
```suggestion
fn make_cgl_conetxt_current(cgl_context: CGLContextObj) -> Result<(), Error> {
```

mrobinson (changes requested):
Great stuff. I just have a few comments. In addition to what's below, please squash your changes into a single commit. Thank you!

.gitignore
sagudev: I will remove this.

src/platform/generic/egl/surface.rs
sagudev: Actually to be compatible with old code it should be `.ok()` instead of unwrap.

src/platform/egl/surface/android_surface.rs
sagudev: There is no gleam in surfman, but c_uint = u32 (most of the time).

mrobinson: Oh, of course. Thanks.

src/platform/macos/cgl/context.rs
sagudev: This won't work as we use err later.

sagudev:
> Great stuff. I just have a few comments. In addition to what's below, please squash your changes into a single commit. Thank you!

I will squash them before adding to queue.

mrobinson (approved):

PR #323, opened 2024-11-23
Cleanup matrix and run tests in CI

sagudev:
I discovered in https://github.com/servo/surfman/pull/321 that we do not run any test in CI, so I cleaned up matrix, run tests on native platforms (use wayland/x11 emulation on linux) and fix examples (mostly by removing support for rwh05 that wasn't working anyway). I think it should be safe to remove rwh05 as winit only supports rwh06, but that should be done as a followup.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,74 +13,77 @@ env:
 
 jobs:
   Build:
-    runs-on: ${{ matrix.os }}
+    name: ${{ format('{0} {1}', matrix.platform.target, matrix.features)}}
+    runs-on: ${{ matrix.platform.os }}
+    env:
+      RUST_BACKTRACE: 1
     strategy:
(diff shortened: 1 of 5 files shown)


jdm (approved):

sagudev:
I've added https://github.com/servo/surfman/pull/323/commits/f7280b645be9ae7e7dd2a36023103c50a7c8453b that ignores some bad results from angle.

BTW I do not have permissions to land in surfman :(

jdm:
I just fixed the repo settings; the Maintainers team only had read access for some reason.

sagudev:
> I just fixed the repo settings; the Maintainers team only had read access for some reason.

I still have: slika

jdm:
I have now added the maintainers team to the list of authorized users.

sagudev:
It works!

Issue #325, opened 2024-12-17
New surfman release


jschwe:
Can we get a new surfman release, so we can switch back from surfman as a git dependency to a normal crates.io dependency?

sagudev:
My plan is to also get https://github.com/servo/surfman/pull/321, because it's also breaking.

sagudev:
Update: #321 landed, but I plan to do release on weekend (and before, land companion PR to servo).

mrobinson:
Is there anything special necessary for the release? I'm happy to do this sooner than the weekend, if there aren't any complications.

sagudev:
I was hoping to update some docs, but I will not have time for it during the week.

jschwe:
@sagudev Is there still something blocking a new surfman release?

sagudev:
No, at least not from my side anymore.

atouchet:
It would be nice to get this done. @mrobinson are you still able to put out a new release?

atouchet:
This was done: https://crates.io/crates/surfman/0.10.0

PR #326, opened 2024-12-30
fix(macOS): remove layer from superlayer when dropping

wusyong:
https://github.com/servo/servo/pull/34780 's macOS smoketest failed because of double free on CALayer.
I totally forgot I had to deal with various double free issues on macOS when I attempted to make window's lifetime shorter.
Winit window will also try to free it when dropping.

Could we make a new release after this?


sagudev:
> Could we make a new release after this?

You can just pin servo to new version, see https://github.com/servo/surfman/issues/325.

wusyong:
Ah you are right. I'll update `rev` later on.

jdm (approved):

PR #330, opened 2025-01-22
Update documentation to reflect when a context is made current and when it needs to be made current

sagudev:
as promised more docs, there are still some functions that require context to be current and I would like to document them too.


mrobinson (approved):

sagudev:
> there are still some functions that require context to be current and I would like to document them too

I do not remember which (I've been cooking this too long anyway) so lets just merge this as is.

PR #331, opened 2025-02-02
Add android CI and fix android

sagudev:
We need to ensure that EGL context is current.

Discovered in https://github.com/servo/servo/pull/34328

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,13 +10,12 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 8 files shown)


sagudev:
I think this is mostly done,
`test_newly_created_contexts_are_current` failes because we use expect no surface is set to context, but in android context cannot be made current if no surface attached, so we use dummy surfaces hence we fail this test. I think we can remove that part of testing.

as for `test_context_creation` I think the problem is same as in mozangle

I will update expectations tomorrow.

jschwe (approved):

Issue #329, opened 2025-01-07
Surfman tests freeze on "chains sm-no-wgl sm-angle-builtin"


sagudev:
runs:
https://github.com/servo/surfman/actions/runs/12647771161/job/35240794466
https://github.com/servo/surfman/actions/runs/12630057884/job/35189122571

also happens on main

I suspect there might be mutex deadlock.

sagudev:
One thread (test_context_sharing) is at https://github.com/servo/surfman/blob/300789ddbda45c89e9165c31118bf1c4c07f89f6/src/platform/windows/angle/device.rs#L81 other are waiting on test serial

sagudev:
using 1 test threads doesn't work, but running test one by one (manually) works so I guess test are not really isolated.

sagudev:
I am wondering if removing serial and only limiting test threads would work.

sagudev:
> I am wondering if removing serial and only limiting test threads would work.

It wont. Actually using `--test-threads 1` will always freeze. Maybe we should use https://github.com/nextest-rs/nextest

sagudev:
I think the real problem is tear down (dropping) that poisons something.

mrobinson:
I've been doing a lot of investigation into this issue and have tried a couple things: (#365, #366). Neither of these have worked and disabling Windows Defender doesn't seem to help either. What I do know is that the test is freezing the second time a test calls `CreateDXGIFactory1`.

mrobinson:
I've tried switching to `cargo-nextest` here: #367. If nothing else this should make the tests run a bit faster.

PR #333, opened 2025-03-11
Switch to x11-dl

Ella-0:
Everything else in servo uses x11-dl and this allows building on systems without X11 libraries installed.


jdm (approved):
Makes sense!

Ella-0:
Forgot rustfmt. Should be fixed now.

sagudev:
CI problem is https://github.com/servo/surfman/issues/329

PR #334, opened 2025-04-20
Use `objc2-*` family of crates

madsmtm:
The `objc2-*` family of crates provide many benefits over the `core-foundation-rs` crates, most notably here is that the bindings are auto-generated and thus much more complete (meaning that we can avoid the error-prone `msg_send!` in almost all cases).

See https://github.com/servo/core-foundation-rs/issues/729 for additional motivation.

### DIFF
--- a/src/platform/macos/cgl/connection.rs
+++ b/src/platform/macos/cgl/connection.rs
@@ -125,15 +125,27 @@ impl Connection {
         raw_handle: rwh_05::RawWindowHandle,
         _size: Size2D<i32>,
     ) -> Result<NativeWidget, Error> {
-        use crate::platform::macos::system::surface::NSView;
-        use cocoa::base::id;
+        use objc2::{rc::Retained, MainThreadMarker};
+        use objc2_app_kit::{NSView, NSWindow};
         use rwh_05::RawWindowHandle::AppKit;
 
         match raw_handle {
-            AppKit(handle) => Ok(NativeWidget {
-                view: NSView(unsafe { msg_send![handle.ns_view as id, retain] }),
-                opaque: unsafe { msg_send![handle.ns_window as id, isOpaque] },
-            }),
+            AppKit(handle) => {
+                assert!(
+                    MainThreadMarker::new().is_some(),
+                    "NSView is only usable on the main thread"
+                );
+                // SAFETY: The pointer is valid for as long as the handle is,
+                // and we just checked that we're on the main thread.
+                let ns_view = unsafe { Retained::retain(handle.ns_view.cast::<NSView>()).unwrap() };
+                let ns_window =
+                    unsafe { Retained::retain(handle.ns_window.cast::<NSWindow>()).unwrap() };
+
+                Ok(NativeWidget {
+                    view: ns_view,
+                    opaque: unsafe { ns_window.isOpaque() },
+                })
+            }
             _ => Err(Error::IncompatibleNativeWidget),
         }
     }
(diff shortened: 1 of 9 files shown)


src/platform/macos/cgl/surface.rs
```diff
@@ -76,6 +77,38 @@ impl Debug for SurfaceTexture {
...
+            0,
+        );
+
+        if gl_error != kCGLNoError {
+            let error_msg = CStr::from_ptr(CGLErrorString(gl_error));
+            panic!("{}", error_msg.to_string_lossy());
+        }
+    }
+}
+
```
madsmtm, line 111: This was previously provided in [`io_surface::IOSurface::bind_to_gl_texture`](https://docs.rs/io-surface/0.16.0/io_surface/struct.IOSurface.html#method.bind_to_gl_texture), but it really mixes two domains (IOSurface and CGL), and requires that we link both of those, so I didn't want to expose it in `objc2-io-surface`.

src/platform/macos/system/ffi.rs
```diff
@@ -4,35 +4,8 @@
...
-
-pub(crate) const kCVPixelFormatType_32BGRA: i32 = 0x42475241; // 'BGRA'
-
-pub(crate) const kCVReturnSuccess: i32 = 0;
-
 pub(crate) const kIODefaultCache: i32 = 0;
 pub(crate) const kIOWriteCombineCache: i32 = 4;
 pub(crate) const kIOMapCacheShift: i32 = 8;
 pub(crate) const kIOMapDefaultCache: i32 = kIODefaultCache << kIOMapCacheShift;
 pub(crate) const kIOMapWriteCombineCache: i32 = kIOWriteCombineCache << kIOMapCacheShift;
```
madsmtm, line 11: Provided in `objc2-io-kit`, but felt like too little benefit to take on the dependency.

src/platform/macos/system/surface.rs
```diff
@@ -461,16 +492,9 @@ impl<'a> Drop for SurfaceDataGuard<'a> {
...
-    #[inline]
-    fn drop(&mut self) {
-        unsafe {
-            let () = msg_send![self.view.0, release];
+            self.surface
+                .io_surface
+                .unlock(IOSurfaceLockOptions::empty(), &mut seed);
         }
     }
 }
```
madsmtm, line 512: This kind of manual memory management is now unnecessary, the `Retained<NSView>` in `NativeWidget` will do it for you.

src/platform/macos/system/surface.rs
```diff
@@ -123,8 +122,8 @@ impl Device {
             let size = match surface_type {
                 SurfaceType::Generic { size } => size,
                 SurfaceType::Widget { ref native_widget } => {
-                    let window: id = msg_send![native_widget.view.0, window];
-                    let bounds = window.convertRectToBacking(native_widget.view.0.bounds());
+                    let window = native_widget.view.window().unwrap();
```
mrobinson: Maybe it makes sense to use `expect(...)` here as you do in places where you get the window from an NSView?

src/platform/macos/cgl/connection.rs
```diff
@@ -125,15 +125,27 @@ impl Connection {
...
+            AppKit(handle) => {
+                assert!(
+                    MainThreadMarker::new().is_some(),
+                    "NSView is only usable on the main thread"
+                );
+                // SAFETY: The pointer is valid for as long as the handle is,
+                // and we just checked that we're on the main thread.
+                let ns_view = unsafe { Retained::retain(handle.ns_view.cast::<NSView>()).unwrap() };
+                let ns_window =
+                    unsafe { Retained::retain(handle.ns_window.cast::<NSWindow>()).unwrap() };
```
mrobinson: Is it important to explicitly retain the window here if you are simply calling `isOpaque()` and discarding the smart pointer right after?

mrobinson (commented):
Looks great! Just a couple questions here:

src/platform/macos/cgl/connection.rs
madsmtm: Nope, it isn't, I've changed it

mrobinson (approved):

madsmtm:
I've marked this as ready now, since https://github.com/servo/core-foundation-rs/issues/729 seems to be settling on deprecating the `core-foundation-rs` crates.

Unsure why CI failed, maybe spurious?

PR #335, opened 2025-05-18
Update wayland-sys to 0.31

webbeef:
This will allow servo to get rid of the `wayland-sys` duplicate.


webbeef:
Not sure what's happening with the failing X11 test - it's passing locally.

jdm (approved):

webbeef:
@jdm is there any chance the test failure is intermittent? I tried locally both in a Wayland and in an X11 session and could not reproduce it.

jdm:
It appears the answer is yes, and it also appears that we have a preexisting timeout issue with several tests on windows :(

PR #350, opened 2025-11-13
Disable `test_cross_device_surface_texture_blit_framebuffer`

mrobinson:


jdm (approved):

mrobinson:
Sadly it seems like other tests are timing out. The situation with the CI is quite bad lately. :/

sagudev:
windows problem is https://github.com/servo/surfman/issues/329, but CI was always bad in surfman. I think nextest would help us (and allow us to remove serial).

PR #351, opened 2026-04-08
feat: Add method to access DXGI share handle

burhankhanzada:
**Summary**
Add a public `share_handle()` accessor to `Surface` on the Windows ANGLE and WGL backends, returning the underlying DXGI shared handle (`HANDLE`) for generic (pbuffer/texture) surfaces.

**Motivation**
Consumers that need to share a surfman `Surface` with other GPU APIs (e.g. importing into a D3D12/wgpu device via `OpenSharedHandle`) currently have no way to obtain the DXGI share handle through the public API. The handle is already computed and stored internally during surface creation, but it is gated behind `pub(crate)` fields.

This is needed for DirectX rendering bridge, where ANGLE renders into a surfman surface and the backing D3D11 texture is then shared with wgpu's DX12 backend for compositing.

### DIFF
--- a/src/platform/windows/angle/surface.rs
+++ b/src/platform/windows/angle/surface.rs
@@ -599,6 +599,15 @@ impl Surface {
             _ => false,
         }
     }
+
+    /// Returns the DXGI share handle if it has one.
+    #[inline]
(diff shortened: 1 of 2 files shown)


jdm (approved):
Makes sense!

PR #360, opened 2026-06-03
Only call `eglMakeCurrent` if anything has changed in `CurrentContextGuard::drop`

mrobinson:
Instead of unconditionally calling `eglMakeCurrent` when restoring the
context in `CurrentContextGuard::drop` only do this if anything has
changed. This guards against drivers that do too much when calling
`eglMakeCurrent`.

### DIFF
--- a/src/platform/generic/egl/context.rs
+++ b/src/platform/generic/egl/context.rs
@@ -62,27 +62,51 @@ pub(crate) struct CurrentContextGuard {
     old_egl_context: EGLContext,
 }
 
-impl Drop for EGLBackedContext {
-    #[inline]
-    fn drop(&mut self) {
-        if self.egl_context != egl::NO_CONTEXT && !thread::panicking() {
-            panic!("Contexts must be destroyed explicitly with `destroy_context`!")
-        }
+impl CurrentContextGuard {
+    pub(crate) fn new() -> CurrentContextGuard {
+        EGL_FUNCTIONS.with(|egl| unsafe {
+            CurrentContextGuard {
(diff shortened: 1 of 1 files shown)


src/platform/generic/egl/context.rs
```diff
@@ -62,30 +62,54 @@ pub(crate) struct CurrentContextGuard {
...
-                    self.old_egl_read_surface,
-                    self.old_egl_context,
-                );
+            if self.egl_display == egl::NO_DISPLAY {
+                return;
             }
+
+            // Only call eglMakeCurrent again if anything has changed.
+            if egl.GetCurrentDisplay() == self.egl_display
+                && egl.GetCurrentSurface(egl::DRAW as EGLint) != self.old_egl_draw_surface
```
jdm: Why are we using != for two if these checks?

jdm:
What is the effect of these changes?

src/platform/generic/egl/context.rs
mrobinson: Oops. I accidentally uploaded an old version of this change. This should be addressed now.

mrobinson:
> What is the effect of these changes?

In the end I think this depends on what the driver does when you call `eglMakeCurrent` with a set of arguments that represents the current state. I created this change as part of tracking down spurious calls to `eglMakeCurrent` while fixing https://github.com/servo/servo/issues/30827. In the end this change didn't fix the issue, but it seems that we should avoid calling `eglMakeCurrent` again if we don't actually need to, as we are then leaving it up to the driver to be well-behaved.

jdm (approved):

PR #362, opened 2026-06-07
Add a `Device::resize_bound_surface` API

mrobinson:
Much like #361, this change adds a `Device::resize_bound_surface` API so
that widget surfaces can be resized without introducing a spurious
`eglMakeCurrent` with `NO_SURFACE`. This will greatly improve resizing
behavior in Servo.

### DIFF
--- a/src/platform/macos/cgl/surface.rs
+++ b/src/platform/macos/cgl/surface.rs
@@ -8,7 +8,10 @@ use crate::context::ContextID;
 use crate::gl_utils;
 use crate::platform::macos::system::surface::Surface as SystemSurface;
 use crate::renderbuffers::Renderbuffers;
-use crate::{gl, Error, SurfaceAccess, SurfaceID, SurfaceInfo, SurfaceType, WindowingApiError};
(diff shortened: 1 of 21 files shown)


src/platform/macos/cgl/surface.rs
```diff
@@ -354,15 +357,23 @@ impl Device {
...
-
-        self.0.resize_surface(&mut surface.system_surface, size)?;
-
         let context_descriptor = self.context_descriptor(context);
         let context_attributes = self.context_descriptor_attributes(&context_descriptor);
+        self.resize_inner(surface, size, &context.gl, context_attributes)
+    }
+
+    /// Resizes a widget surface.
+    pub fn resize_inner(
```
jdm: Should this be pub(crate)?

jdm (approved):

src/platform/macos/cgl/surface.rs
mrobinson: Yep! I will correct that before landing.

PR #365, opened 2026-06-07
Remove duplicate `eglDestroyContext` call in angle `Device::destroy_context`

mrobinson:
It's quite likely that this duplicate `eglDestroyContext` call is
causing the Windows CI to hang. This change removes the duplicate call
to context destruction.


mrobinson:
The CI still seems to be hanging, but this is still a good fix, I think.

jdm (approved):
Super weird this was here at all!

mrobinson:
> Super weird this was here at all!

Agreed. I tracked this down to the initial implementation, so I suspect it was just a copy-and-paste issue or something along those lines.

Issue #327, opened 2024-12-30
Too much code duplication


sagudev:
There is a lot of code duplication in surfman egl that could be avoided if we used traits and default implementations.

mrobinson:
I think this crate could be a lot simpler if instead of having only a single implementation of every `Surface` and `Device`, each `Surface` and `Device` was internally an enum of the different backends -- the same way that `raw_window_handle`, etc are designed. That would also allow using more code reuse, I think.

PR #369, opened 2026-06-08
Rework project directory structure

mrobinson:
This is a major simplification of the project directory structure. An
example of what was confusing before was that there were multiple
directories named `generic` and `egl`. Now modules are named more
specifically and the directory structure is flattened. The new
structure:

`src/`
 - `angle/` ← *Windows ANGLE backend*
 - `cgl/` ← *macOS CGL backend*
 - `hardware_buffer/` ← *Android / OHOS hardware buffer backends*
 - `mesa_surfaceless/` ← *Mesa-specific pbuffer surface backend*
 - `multi/` ← *multi backend*
 - `wayland/` ← *Wayland backend*
 - `wgl/` ← *Windows WGL backend*
 - `x11/` ← *X11 backend*
 - `base` ← *a directory for code that can be shared by multiple backends*
   - `io_surface` ← *IOSurface shared code (could be used by future Metal backend)*
   - `egl` ← *shared EGL types*

In addition, this commit removes the full file path included as a
comment at the top of every source file. These paths are now incorrect
and they were likely only there because the directory structure was so
confusing.

This is part of #327.


jdm (approved):

jdm:
Good idea!

src/hardware_buffer/mod.rs
```diff
@@ -1,5 +1,3 @@
-// surfman/surfman/src/platform/egl/mod.rs
-//
 //! Bindings to EGL on Android.
```
mukilan, line 1: I guess this comment needs to be updated. I guess this was misleading even before this change though.

PR #371, opened 2026-06-09
Merge most `impl Device` blocks

mrobinson:
This change just moves code and adjust imports.  It was previously
difficult to know where any particular `Device` method was implemented.
This will also make it easier to eliminate `macros.rs`, necessary to
implement the new version of the API. The downside is that now much of
the implementation is gathered into the `device.rs` files, but these
usually do not go beyond 1500 lines. Further changes to the API will
eliminate many lines of code and spread the logic more equally between
files.

For the moment, the code in the `hardware_buffers` `Surface` is mostly
untouched. This needs more careful surgery.


mrobinson:
I realized this is likely very difficult to review so I've split this up into 6 separate PRs in order to make this a bit easier:

- #372 
- #373 
- #374 
- #375 
- #376 
- #377 
- #378 

Just a word on my motivation as I know this looks like a lot of useless code removal. In order to make any progress at all on making surfman easier to work on, we need to remove the giant macro in `macros.rs` which interferes with making API changes as well as using rust-analyzer to work on things. In order to do that all of the `impl Device` code needs to be in the same place in order to implement a trait properly. In addition, it was very difficult to know before where to look for the implementation of any `Device` method as they were scattered throughout many files

Issue #370, opened 2026-06-08
Revamped API Proposal


mrobinson:
In the interest of simplifying the `surfman` and preparing for adding support for interoperability with `wgpu` and for cross-process surface handles, I would like to propose version 2 of the API:

 
## Connection

```rust
pub enum Connection {
    Android(...),
    Angle(...),
    Cgl(...),
    Ohos(...),
    SurfacelessMesa(...),
    Wayland(...),
    Wgl(...),
    X11(...),
}
```

## Adapter

```rust
pub enum Adapter {
    Android(...),
    Angle(...),
    Cgl(...),
    Egl(...),
    Ohos(...),
    Wgl(...),
}
```

## Surface

```rust
pub enum Surface {
    Window(WindowSurface),
    Buffer(BufferSurface),
}

pub enum BufferSurface {
    Android(...),
    Direct3d(...),
    IoSurface(...)
    Ohos(...),
    Pbuffer(...),
}

pub enum WindowSurface {
    Android(...),
    AppKit(...)
    Ohos(...),
    Wayland(...),
    Win32(...),
    X11(...),
}
```

Notes:
  - Only `WindowSurface` would implement `present` meaning it's not possible to call present on a buffer surface (this causes an error today).
  - Mutability concerns would be handled by a mutable getter for the currently bound surface:
    ```rust
    if let Some(Surface::Window(window_surface)) = cnnection.bound_surface_mut() {
        window_surface.present(&device);
    }
    ```
 - `Multi` is eliminated. Instead a factory returns the most appropriate context given what is enabled at compile-time and what is currently running (when choosing between X11 and Wayland).
 - This opens up the possibility of other kinds of offscreen buffers such as GBM, which will be necessary for cross-process rendering and integration with `wgpu`.
 - The idea is that `BufferSurface` will provide a fallible method to get a shareable handle to be used for cross-process surfaces.

jdm:
this makes sense! Where do platform-specific extensions like https://github.com/servo/surfman/blob/9eaf3fc40361d1e66a1fc0ef07c6a02bebb05fc0/src/angle/surface.rs#L422 fit into this new model?

mrobinson:
I think there are a couple things we could do here.

1. For some APIs such as `create_surface_from_texture` used on ANGLE, I think it will make sense to have some kind of external handle API for buffer sharing (with `wgpu` and to send across the process boundary). In those cases the handle could also be used to import. Image:
```rust
 pub enum BufferHandle {
[...]

PR #385, opened 2026-09-16
Make `Adapter` an enum

mrobinson:
This is part of #370. It makes `Adapter` an enum that is valid on all
platforms and moves all adapter code to separate files. This already
begins the removal of the code in `Multi`.


src/hardware_buffer/adapter.rs
```diff
@@ -0,0 +1,11 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/. */
+
+//! A hardware display adapter on Android and OHOS systems.
+
+/// Represents a hardware display adapter that can be used for rendering (including the CPU).
+///
+/// Adapters can be sent between threads. To render with an adapter, open a thread-local `Device`.
```
sagudev: I think we should just link to the Adapter docs here and keep only platform specific notes here, instead of duplication the docs and risking going out of sync.

And similarly in all other places.

mrobinson: Okay. I've done this now in a new commit.

src/angle/adapter.rs
```diff
@@ -21,9 +21,7 @@ thread_local! {
     static DXGI_FACTORY: RefCell<Option<ComPtr<IDXGIFactory1>>> = RefCell::new(None);
 }
 
-/// Represents a hardware display adapter that can be used for rendering (including the CPU).
-///
-/// Adapters can be sent between threads. To render with an adapter, open a thread-local `Device`.
+/// An implementation of [`crate::Adapter`] for ANGLE.
```
sagudev: `for ANGLE` -> `for ANGLE on Windows`

Mozilla will switch to metal via ANGLE on macos, as ANGLE will not support GL on mac.

sagudev (approved):

src/angle/adapter.rs
mrobinson: I did not know this. Fascinating. Thank you for the review!

PR #387, opened 2026-09-16
 Remove support for version 0.5.x of `raw-window-handle`

mrobinson:
As we are going to break API, it make sense to stop supporting this very
old version of `raw-window-handle`.


mrobinson:
Draft as this is based on #386.

mrobinson:
This is ready for review now.

sagudev (approved):

.github/workflows/main.yml
```diff
@@ -32,11 +32,10 @@ jobs:
         features:
           [
             "",
-            "chains sm-raw-window-handle-06",
-            "chains sm-raw-window-handle-05",
+            "chains sm-raw-window-handle",
           ]
         include:
-          - features: "chains sm-raw-window-handle-06"
+          - features: "chains aw-window-handle"
```
sagudev:
```suggestion
          - features: "chains raw-window-handle"
```

mrobinson: 🤦

PR #388, opened 2026-09-16
Only run a single macOS job in CI

mrobinson:
It's too difficult to schedule a macOS runner for this project and
scheduling a single job instead of 3 should increase the chances of
success. This just schedules the most common configuration on this
runner.


sagudev (approved):
I just planned to comment on macos usage. Maybe we should switch surfman to rebase and squash model like we did with mozjs: https://github.com/servo/mozjs/issues/794

mrobinson:
> I just planned to comment on macos usage. Maybe we should switch surfman to rebase and squash model like we did with mozjs: [servo/mozjs#794](https://github.com/servo/mozjs/issues/794)

I think we could go a long way by greatly reducing the number of jobs. #387 will cut down a lot on jobs and we could also consider just making `raw-window-handle` mandatory (depending on the platform) -- which I think is probably fine.

PR #390, opened 2026-09-18
Remove the `sm-raw-window-handle` and `chains` features

mrobinson:
This code is now compiled by default. This adds a permanents dependency
on both `fnv` and `raw-window-handle`, which is probably not a big deal.
surfman doesn't have many consumers and they likely already depend on
these things.

One of the big motivations here is to reduce the number of jobs on CI.


jdm (approved):

.github/workflows/main.yml
```diff
@@ -29,25 +29,18 @@ jobs:
               test: false,
             }
           - { target: arm-linux-androideabi, os: ubuntu-24.04, test: false }
-        features:
-          [
```
sagudev, line 33: can we keep explicit `features: [""]` in the matrix for clarity

sagudev (approved):

PR #392, opened 2026-09-18
Release version 0.14

mrobinson:
The idea with this release is that this will include all fixes before
the major API breakages that will be coming for #370.


sagudev:
So this should include all current open PRs except https://github.com/servo/surfman/pull/385?

mrobinson:
> So this should include all current open PRs except #385?

That is one way to do it, I think.

sagudev (approved):

sagudev:
I raised limits for timeout to 120min, but I still suggest we try to land this later as CI is currently very busy in servo/servo.

PR #393, opened 2026-09-19
Collapse various adapter construction methods into one that takes a preference

mrobinson:
This removes a lot of code, matches more closely how `wgpu` creates
creates adapters, and will make it easier to make further changes to all
of these interfaces.


src/adapter.rs
```diff
@@ -16,6 +16,36 @@ use crate::hardware_buffer::adapter::HardwareBufferAdapter;
...
 use crate::Error;
 
+/// A power usage preference for selecting an adapter.
+#[derive(Copy, Clone, Debug, Default)]
+pub enum PowerPreference {
+    /// Prefer a high-power adapter.
+    #[default]
+    High,
+    /// Prefer a low-power adapter.
+    Low,
```
sagudev: Can we have more descriptive names here (like we did for WGL) and [like webgpu
](https://www.w3.org/TR/webgpu/#enumdef-gpupowerpreference)
```suggestion
    /// Prefer a high-power adapter.
    #[default]
    HighPerformance,
    /// Prefer a low-power adapter.
    LowPower,
```

sagudev (approved):

src/adapter.rs
mrobinson: Sure! That's done now.

PR #395, opened 2026-09-20
Eliminate `NativeWidget` entirely

mrobinson:
Instead of a temporary `NativeWidget` type just have the surface
constructor take a `SurfaceType` that contains a `WindowHandle`. This is
now possible since `raw_window_handle` is an unconditional dependency.
This allows removing hundreds of lines of boilerplate.


jdm (approved):
Excellent.

